PR #18, opened 2015-04-02
Partner fixes

coop182:
Hey,

Finally had time to sit down at get this library going for partner applications.

The major road block in the whole thing seems to have been that the signature was being escaped, not sure how this has been working previously but I cannot get it working with the escaping in place.

Also put the concept of refreshing a token in... it feels like this (and maybe even the other requests) should be potentially be simplified away a little more into methods on the Application rather than working directly with the URL and Request classes... perhaps you have some ideas around this?

Anyway let me know what you think and any suggestions or changes you might have... this should now be usable for us in our application so we will look to transition on to this library and help you flesh it out some more as we run into any issues.

Thanks for the effort of getting this going!

Cheers,

Matt

### DIFF
--- a/examples/partner.php
+++ b/examples/partner.php
@@ -26,10 +26,13 @@ $config = array(
 
 $xero = new PartnerApplication($config);
 
-//if no session or if it is expired
-if(null === $oauth_session = getOAuthSession()){
+$oauth_session = getOAuthSession();
+
+// If no session found
+if($oauth_session === null){
 
     $url = new URL($xero, URL::OAUTH_REQUEST_TOKEN);
+
     $request = new Request($xero, $url);
 
     //Here's where you'll see if your keys are valid.  You can catch a BadRequestException.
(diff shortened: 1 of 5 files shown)


calcinai:
Awesome, thanks for your work on this.

I'm just testing with my existing test apps to make sure there are no negative implications from the changes.  I like the idea of wrapping some of the functionality up in the Application class itself.

As far as not escaping of the signature goes, your branch now crashes IIS at Xero's end for my test private app.  I can only assume that this is because you sign using a query string parameter, not a header?  Hopefully this is the case, as I can see that this would escape it twice.  All this would mean, is that we needed to add the escaping on ```Remote\Request:85```.

One of these things would be moving the ```->getAuthorizeURL()``` function into the app, as it's static as far as the app is concerned, and you wouldn't need to pass any arguments in, it could simply return a ```new URL()``` pre-populated with the correct base url and path.  Then, for consistency, you could use the existing ```->getFullURL()``` method on it.  Otherwise it could probably just return the string directly.

Again, thanks for your help with this.  Let me know how you want to proceed - I'm happy to make those changes and test at my end if you like.

coop182:
Yeah we sign in the query string - just tested with the header... and yeah its not pretty - just had a little go at fixing it but no luck yet.

calcinai:
```$header = 'OAuth ' . Helpers::flattenAssocArray($oauth_params, '%s="%s"', ', ', null, true);```

Should do the trick on that one.  Want me to update your PR?

coop182:
Yeah go for it!

PR #19, opened 2015-04-02
Commonledger partner fixes

calcinai:
Minor alterations to PR #18


calcinai:
@coop182 - Do you want to check out this branch and test?  The only difference is that ```getAuthorizeURL()``` method that I added to the app.

If this is all ok, I'll maybe try to bundle some of the other calls up for simplicity.

coop182:
@calcinai - Just gave it a test and it all seems to be working really well :smile: 

Will start using this properly next week and will send through any other changes we think will be useful.

PR #20, opened 2015-04-07
Add missing query options

coop182:
Adding missing query options for when you are working with collections of objects...

I think this is the right way to do it? I explored adding optional parameters to each model that has them through parsing the pages with the generator... but that got complicated fast... so hopefully this is all good haha!

### DIFF
--- a/src/XeroPHP/Remote/Query.php
+++ b/src/XeroPHP/Remote/Query.php
@@ -12,11 +12,17 @@ class Query {
     private $from;
     private $where;
     private $order;
+    private $modifiedAfter;
+    private $page;
+    private $offset;
 
     public function __construct($app) {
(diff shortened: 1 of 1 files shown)


calcinai:
Looks reasonable to me - if you looked in the generator, you might have seen me catching the block which matches if the model supports page/offset, but in reality it's probably not worth it!  

It's not going to have any negative consequences if you add one of those parameters to a request that doesn't support it, ultimately it's up to the user to know if it will have any effect.

coop182:
Oh yeah I see it now... yeah I like the idea of having it in the Query but I will take a look at this again when I get a bit more time and see if it would actually be better to do it in each Model...

For now this lets us get cracking :smile:

PR #21, opened 2015-04-10
Small changes

coop182:
Just a few readme tweaks.


calcinai:
Cheers for doing that.  Xero gave me a partner app for testing yesterday, so I can actually test (rather than just blindly assume it'll work)

coop182:
Awesome, as of yesterday we are now running this library in production as well... going great so far :smile:

Issue #4, opened 2015-03-09
Use Guzzle for OAuth? Drop PHP 5.3 support


nfauchelle:
Hello,

Are you pretty set that this must support 5.3?

I was thinking maybe the oauth stuff could be swapped out for Guzzle / https://github.com/guzzle/oauth-subscriber 

But that would need 5.4+

Nick

paulsschwarz:
+1 for Guzzle

calcinai:
I did consider using Guzzle (it's actually used in the model generator), but it just seemed a little heavy for what was required (unless of course, the project already used it).  My thinking was to make it pluggable, so you could use Guzzle if it was already in your project, or the internal lightweight one in other cases.  Didn't quite get that far!

nfauchelle:
So.. that means your fine with 5.4 then? :)

calcinai:
That decision wasn't so much about me - anything that I have anything to do with is on  PHP >= 5.4, I just didn't want to put anything in the way of people wanting to move off that one-class-monolithic-monster.

I have no objection if it's for the greater good, but as I said earlier, It'd be nice if the user had the option.

phindmarsh:
I think the general feeling would be if you are just going to implement pluggable transports it seems like you could easily keep the `5.3` compatibility with the existing implementation, but require `>=5.4` when using (say) the Guzzle transport instead. That way you don't shut out anyone (unfortunate enough to be) stuck on a version that Guzzle doesn't support.

Guzzle can be a little like using a sledgehammer to crack a walnut for small projects, but if it is already included in a project it makes complete sense to leverage it. IMO, having options is really useful.

drewbroadley:
I agree with @phindmarsh on this, mainly around the fact this library tries to build its own OAuth client, which can be questionably as bad as including a static copy of an OAuth library rather than having it as a composer dependancy.

Now the main reason I agree with @phindmarsh is because guzzle has an extension to then handle OAuth1.0 (https://github.com/guzzle/oauth-subscriber) and the RSA-SH1 signing.

We are currently battling this exact issue with the library and using a partner programme. We will commit pull requests to make this work, but I think it needs to head to using something like guzzle + OAuth 1.0a library which can handle RSA-SH1 signing.

I'm tossing up if we fork to use Guzzle and the OAuth subscriber extension.

nfauchelle:
@drewbroadley 
I agree on using that oauth-subscriber (mentioned in the OP).

[...]

PR #23, opened 2015-05-05
Added basic events class and triggers for request and response

coop182:
We have a requirement to be able to get at the raw request and response objects to enable us to log them and use them for debugging in the event of an error.

Using the original Xero library this is pretty easy because you have direct access to the request method and can basically log what you send and receive from it... this Library has abstracted the raw request method away (which is a good thing!) and so its now no longer possible to do what we need... at least I think thats the case...

I am proposing the following as a solution but I am very open to other suggestions - it is just one of many possible solutions I am sure and may not fit in at all with where you want to take this Library...

So yeah feel free to tell me its a crap idea... or let me know how you would rather it is done...

With what I have done it is now possible to listen for events like this:

```php
<?php 

use XeroPHP\Application\PartnerApplication;
use XeroPHP\Remote\Request;
use XeroPHP\Remote\Response;

$xero = new PartnerApplication($config);

// Request event
$xero->events()->listen('onRequest', function(Request $request) {
	echo $request->getBody();
});

// Response event
$xero->events()->listen('onResponse', function(Response $response) {
	echo $response->getResponseBody();
});

$contact = $xero->loadByGUID('Accounting\\Contact', [GUID]);
```

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -43,11 +43,14 @@ abstract class Application {
 
     protected $config;
     protected $oauth_client;
+    protected $events;
 
     /**
      * @param array $user_config
      */
     public function __construct(array $user_config) {
+        $this->events = new Events();
+
         //better here for overriding
         $this->config = array_replace_recursive(self::$_config_defaults, static::$_type_config_defaults, $user_config);
 
@@ -65,6 +68,12 @@ abstract class Application {
         return $this->oauth_client->getAuthorizeURL();
     }
 
+    /**
+     * @return Events
+     */
+    public function events() {
+        return $this->events;
+    }
 
     /**
      * @param $key
--- a/src/XeroPHP/Events.php
+++ b/src/XeroPHP/Events.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace XeroPHP;
+
+class Events {
+
+    private $_callbacks = array();
+
+    public function listen($event, $callback)
+    {
+        if(!isset($this->_callbacks[$event])) {
+            $this->_callbacks[$event] = array();
+        }
+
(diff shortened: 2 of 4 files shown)


drewbroadley:
To add to this, if this is not ideal I think it it could be another case for adding guzzle. It has its own event driven handlers: http://guzzle.readthedocs.org/en/latest/events.html

calcinai:
Haha, beat me to it!

http://guzzle3.readthedocs.org/http-client/request.html#events-emitted-from-a-request

coop182:
Yeah didn't realise guzzle did that... probably a way better option haha... looks like its guzzle time...

calcinai:
I have a branch where I've started a Guzzle plugin.  I'll try to get it tested this week.

gerrywastaken:
@calcinai Given the difficulty of the Guzzle implementation, I think it would be great to have this rebased and and applied in it's current state. Thoughts?

After Guzzle is added, the event listener could be modified to set up a Guzzle listener instead. This would probably work better if instead of a public events method on the Application, we instead had an addListener method.

This is unless we have very strict backwards compatibility requirements, ie. open closed principle. If so, then it's probably better to just wait.

calcinai:
That's why had left this sitting here - I figured that it'd soon be made redundant.  I'm (again) having second thoughts on Guzzle - what do you think? Check out #4

gerrywastaken:
@calcinai Yup already commented on that thread. I personally dislike working with CURL in PHP. Mostly because debugging it isn't simple: http://curl.haxx.se/mail/curlphp-2005-11/0078.html

I haven't attempted to rewrite the curl code using Guzzle, however it sounds like it might be a long term goal due to the difficulty involved. Where as I think this feature is something that people clearly need, the Guzzle is just a nice to have.

PR #24, opened 2015-05-14
Allow bulk update of Models

coop182:
Small change to `Application->saveAll()` to allow bulk updates... not sure if this is documented with Xero... but it works!

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -196,26 +196,28 @@ abstract class Application {
     public function saveAll(array $objects) {
 
         //Just get one type to compare with, doesn't matter which.
-        $type = get_class(current($objects));
+        $current_object = current($objects);
+        $type = get_class($current_object);
+        $has_guid =  $current_object->hasGUID();
         $object_arrays = array();
-
-        $requestMethod = Request::METHOD_PUT;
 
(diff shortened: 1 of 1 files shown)


coop182:
I have updated it to stop constant reassignment... that was a bit lazy... but I don't want to do the first object comparison and exception if they are different.

If I do that it either has to be a bulk add or update... the way I have it now it can add and update.

That's what I saying about it not being documented by Xero as being possible... but it works... if you send a mixed array of objects some with and without GUID's it will update the ones with GUID's and new records for the ones without. I have successfully tested this against the ManualJournals endpoint.

coop182:
Actually looking at the docs... it is clearly supported to create and update via POST... its actually kinda tempting to just change this to POST and take everything else out... but there are a few endpoints that only allow create... so I guess this is the best option.

calcinai:
It feels a bit yuck, but I guess no reason not to allow it.  So just change the method on the first one you find with a GUID?

coop182:
Yeah it is a bit yuck... but the benefits of being able to bulk sync data make it worth while I think. So yeah you if we find a GUID in the array of objects switch to POST otherwise PUT.

calcinai:
Sounds good.

PR #25, opened 2015-05-15
Updated generator to attempt to find missing GUID's

coop182:
I really needed the ManualJournalID property added to its model... there's a few others this change has picked up as well... not sure if that is all of them.

### DIFF
--- a/generator/generate
+++ b/generator/generate
@@ -225,7 +225,9 @@ foreach($scrape_apis as $scrape_api) {
 
         $current_model = null;
 
-        $crawler->filter('.entry-content')->children()->each(function ($node) use ($api, &$current_model) {
+        $last = $crawler->filter('.entry-content')->children()->last();
+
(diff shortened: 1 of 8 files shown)


calcinai:
Hopefully Xero will release a set of XSDs that is complete (and covers all APIs) soon, then we can do away with all this scraping business!

I had a thought that I could split this in to 2 projects, one being a scraper/generator of xsd/swagger files, and the other basically everything else. It would probably help moving forward (if we can just point it at a Xero-maintained spec).

coop182:
Yeah that would be the ideal! I take it these are not complete? https://github.com/XeroAPI/XeroAPI-Schemas I haven't looked into them too much.

Scraping the docs is definitely interesting... was looking to the annoying little edge case with this page earlier... http://developer.xero.com/documentation/api/expense-claims/

Generates a sub model when I'm pretty sure it shouldn't... pesky double table in there.

I've heard that a better solution should be on the horizon soon as well... so bring it on!

PR #26, opened 2015-05-18
Generator tweaks

coop182:
The following changes were started because of [this commit](https://github.com/calcinai/xero-php/commit/d154789564f7bea28ac48d9e8de18c7bc3bf43b8) to the Account Model breaking the ability to add an account.

I have tried to tighten things up a bit so that the read only and mandatory properties are set correctly more often.


calcinai:
I tried to make as few assumptions as possible when I was first generating (e.g. mixed case properties), but I think your changes are necessary.

I'm going to have a crack at that separate repo for generation, so let me know directly if you need any changed before I've pushed it out.

We really need to make some tests!

coop182:
I think we have everything in a good state to keep pushing forward with our app now... but will let you know if I hit anymore snags.

As far as breaking the generator code out into its own repo I think thats a great idea... would be nice to keep the crazy confined to a separate place away from the general "niceness" of the rest of the library.

Now that I actually understand the generator code and the state of the docs more, it amazes me that you managed to get the structure out of it you did... its nice work man!

PR #31, opened 2015-06-17
Making InvoiceNumber have the correct type

coop182:
Xero Invoice Numbers are alpha numeric.


coop182:
Yeah cool... agree on both points.

Issue #32, opened 2015-06-17
Uploading attachments


drawmyattention:
Given I have an existing invoice, how would I add an attachment to that invoice?

There is no attachment model, and looking at the ```Save()``` method on the ```Application``` class, there's no way for the file to be saved that I can determine.

drawmyattention:
My issue isn't relating to downloading PDFs, or other existing invoice attachments. I'm referring more to the ability to POST a brand new attachment to the Attachments endpoint as per http://developer.xero.com/documentation/api/attachments/

calcinai:
Yep - I'm about to push a fix for both.

calcinai:
You can use it like this:

```php
$attachment = \XeroPHP\Models\Accounting\Attachment::createFromLocalFile('image.jpg');
$invoice = $xero->loadByGUID('Accounting\\Invoice', '[GUID]');
$invoice->addAttachment($attachment);
```

and fetch them like:
```php
$invoice = $xero->loadByGUID('Accounting\\Invoice', '[GUID]');
$attachments = $invoice->getAttachments();
...
$attachment->getContent(); //Will contain the binary content of the file.
```

Let me know if you fun into any issues.

drawmyattention:
Just checked out the latest version, and having an issue with the file that gets created within Xero.

Using the exact same code that you've posted above, I can see the file gets created in Xero as can see it is listed along with the invoice that I've associated it to. 

However, when I try to view the attachment (via the code above, or even directly within Xero) I get either a blank white document, or 'Failed to load PDF document'.

I've tried three different PDF documents, each from different sources.

calcinai:
Hey,

I've just been doing some testing too, and I'm getting the same thing.  The ```->getContent()``` method seems to contain the correct data, as if I set the content-type header in my test script and echo it, it shows.  I'm just about to do a diff and see where the discrepancy is.

drawmyattention:
Whilst trying to debug what may be going on, I noticed the '?IncludeOnline=true' flag:

> You can set an attachment to be included with the invoice when viewed online (through Xero). This functionality is available for accounts receivable invoices and accounts receivable credit notes. To enable an attachment to be viewed with the online invoice add the parameter IncludeOnline=true to the end of your request.

However, when appending that to the endpoint url the request throws an Exception ```Failed to validate signature```.

The ```addAttachment()``` method sends the request twice, was this intentional?

calcinai:
[...]

PR #33, opened 2015-06-17
Add support for downloading PDFs like invoices by adding an 'Accept' parameter for loadByGUID

FLamparski:
This is a rather quick hack to make downloading PDFs work - maybe it would be better placed as a method of `Invoice` and other classes whose underlying endpoints permit PDF requests. However, it works fine.

To request a PDF Invoice:
```php
$response = $xero->loadByGUID('Accounting\\Invoice', $invoiceId, Request::CONTENT_TYPE_PDF);
$pdfBuffer = $response->getResponse(); // gets the PDF buffer, which you can save to a file or send as a download
```

Since the parameter is optional and has the default of `CONTENT_TYPE_XML`, nothing should break.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -101,29 +101,35 @@ abstract class Application {
 
 
     /**
-     * As you should never have a GUID for a non-existent object, will throw a NotFoundExceptioon
+     * As you should never have a GUID for a non-existent object, will throw a NotFoundException.
+     * You can also add the accept parameter to request PDFs (ie. of invoices)
      *
      * @param $model
      * @param $guid
+     * @param $accept
      * @return mixed
      * @throws Exception
      * @throws Remote\Exception\NotFoundException
      */
-    public function loadByGUID($model, $guid) {
+    public function loadByGUID($model, $guid, $accept = Request::CONTENT_TYPE_XML) {
 
         $class = $this->validateModelClass($model);
 
         $uri = sprintf('%s/%s', $class::getResourceURI(), $guid);
 
         $url = new URL($this, $uri);
-        $request = new Request($this, $url, Request::METHOD_GET);
+        $request = new Request($this, $url, Request::METHOD_GET, $accept);
         $request->send();
 
-        //Return the first (if any) element from the response.
-        foreach($request->getResponse()->getElements() as $element){
-            $object = new $class();
-            $object->fromStringArray($element);
-            return $object;
+        if ($accept === Request::CONTENT_TYPE_XML) {
+          //Return the first (if any) element from the response.
+          foreach($request->getResponse()->getElements() as $element){
+              $object = new $class();
+              $object->fromStringArray($element);
+              return $object;
+          }
+        } else {
+          return $request->getResponse();
         }
 
     }
--- a/src/XeroPHP/Remote/Request.php
+++ b/src/XeroPHP/Remote/Request.php
@@ -40,7 +40,7 @@ class Request {
     private $response;
 
 
-    public function __construct(Application $app, URL $url, $method = self::METHOD_GET) {
+    public function __construct(Application $app, URL $url, $method = self::METHOD_GET, $accept = self::CONTENT_TYPE_XML) {
 
         $this->app = $app;
         $this->url = $url;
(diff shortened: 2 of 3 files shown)


calcinai:
Cheers for that.  I'll have a look at making the model generator add that functionality automatically - #32 has just been opened and exposed the lack of support for files in general.

Only issue with this PR is is that it drops support for fetching as JSON.  Give me a couple of hours to work something out.

calcinai:
What situation would you see this used most?  Would it be acceptable to have a ```getPDF()``` method on the invoice and credit note objects themselves?

I would have assumed that the most common reason for doing this would be to programatically create and invoice, then email it.  Doing it this way would be entirely BC, but it would mean that if you did't have the invoice object to begin with, you'd have to load it (ultimately taking two requests to get the PDF).

FLamparski:
Yes, I think that use case would be the most common one - it was precisely
what prompted me to make this change.

calcinai:
I'd love to just make a trait for this and then could conditionally use it on the objects that support it. I wonder when it'll be safe to drop 5.3..

calcinai:
Hey, sorry about the delay.  I've got this all parsing from the docs again now, and you'll get a ```->getPDF()`` method on objects that support it.

The method will return the raw content of the PDF.

FLamparski:
That sounds great, thanks! Can't wait to test it out!

LucidNinja:
Maybe I'm missing something, but how do you actually convert the response of getPDF() to output a PDF to the browser? Or does this only return the raw content and we still need to build a PDF template?

calcinai:
`getPDF()` will return the binary data for the PDF. You just need to set the `Content-Type` header and stream that to the browser. 

Something like:

```
header('Content-type: application/pdf');
header('Content-disposition: inline; filename="invoice-12345"');
header('content-Transfer-Encoding: binary');
header('Accept-Ranges: bytes');

echo $invoice->getPDF();
```

LucidNinja:
Ok. Is there a better way for me to be retrieving my invoice then. Currently I'm using the load() method. 
`  $invoiceNumber = 'ORC1096';
	$xero = xero_config();

	$pdf = $xero->load(Invoice::class)
  ->where('InvoiceNumber', $invoiceNumber)
  ->execute();

	foreach ($pdf as $doc) {
		$doc->getPDF();
	}

	header('Content-type: application/pdf');
	header('Content-disposition: inline; filename="invoice-12345"');
	header('content-Transfer-Encoding: binary');
	header('Accept-Ranges: bytes');

	echo $doc;
	exit();
`

calcinai:
The above thing is doing nothing with he result of `->getPDF()`, you'll need to echo that.  Also, you'll only be able to send one file per request.

If you're making lots of requests, you could make a request like at the start of this ticket.

LucidNinja:
I only want to do one invoice at a time, but the result of load returns an array? Is there a better way to retrieve one by the Invoice Number? Thanks so much for your fast responses!

LucidNinja:
I've amended my code and the PDF is working, but ideally I'd only return one invoice:

 ```
 $invoiceNumber = 'ORC1096';
	$xero = xero_config();

	$pdf = $xero->load(Invoice::class)
  ->where('InvoiceNumber', $invoiceNumber)
  ->execute();

  header('Content-type: application/pdf');
	header('Content-disposition: inline; filename="invoice-12345"');
	header('content-Transfer-Encoding: binary');
	header('Accept-Ranges: bytes');

	foreach ($pdf as $doc) {
		echo $doc->getPDF();
	}

	exit();
```

**Comment by calcinai** [comment]
I know it's a little weird, but from memory, I think the `Invoices` endpoint accepts an invoice number in place of the guid, so you can probably go 

```
$pdf = $xero->loadByGUID(Invoice::class, $invoiceNumber)->getPDF();

header('Content-type: application/pdf');
header('Content-disposition: inline; filename="invoice-12345"');
header('content-Transfer-Encoding: binary');
header('Accept-Ranges: bytes');

echo $pdf;
```

Exceptions will be thrown if it doesn't exist.

LucidNinja:
Mate - you're a genius! Thanks so much. That works perfectly!

calcinai:
No problem. It might make sense to actually add this to the PDFTrait, so you can just go `$invoice->streamPDF()` or something.  Will have a think.

LucidNinja:
Yeah sounds like a great idea! While I've got you, I've manually pulled in the SendEmailTrait.php from the current master, but this isn't available over Composer yet, because there is no release available with that file. Any ideas when this might be added? Or a better way to include this for now?

PR #35, opened 2015-06-29
Make inaccessable private key exception more informative

gerrywastaken:
Previously, setting
```php
$config['rsa_private_key'] = 'file://full/path/to/key.pem'
```
(note two slashes instead of three) or other variations that are incompatible with openssl_pkey_private(), would just give the user "Cannot access private key for signing". However investigation will lead to confusion as the file is accessible, just not in the correct format. The additional info I've added allows the user to go straight to the manual instead of spending time thinking it's a permissions issue.


calcinai:
I'm sure that'll help some people out!

gerrywastaken:
:+1: Yup already bit me, so I figured others would run into it.

PR #38, opened 2015-07-06
Fix a single character typo

gerrywastaken:


calcinai:
Thanks for that. I think I actually typed up those examples on my phone in the github editor (there's no way my IDE would've let me get away with that!)

gerrywastaken:
Haha, no worries.

side note: Just reading through generate.php atm. That's pretty incredible the way you generated the Remote Objects. Impressive.

PR #39, opened 2015-07-10
Change example for getting journal data to a valid version

gerrywastaken:
As the _data property of remote/object is protected, trying to use json_encode() on it will just return an empty object.

### DIFF
--- a/examples/partner.php
+++ b/examples/partner.php
@@ -101,9 +101,10 @@ if($oauth_session === null){
 // We are in! Grab some journals...
 $journals = $xero->load('Accounting\\Journal')->execute();
 echo sprintf('Found %s journals', count($journals));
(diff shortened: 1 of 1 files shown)


gerrywastaken:
I got caught up on this for quite a while and couldn't work out why my json was showing an empty object. I'm guessing this is why you commented out the code?

calcinai:
I figured it was better to fix the root cause of this.  Since 5.4+ is now a requirement, we can use ```JsonSerializable```

gerrywastaken:
That's even better. Thanks for doing that!

gerrywastaken:
Oh and mentioning toStringArray() in the Readme might be a good idea as it might be useful in other senarios.

PR #41, opened 2015-07-16
Public example / Fixed curl config

mhlavac:
This fixes public application curl config + if there is an error, whole exception is printed for clarity.

Also if response is not available because exception was thrown before response was populated, then it's not printed.


calcinai:
Thanks for that!

mhlavac:
@calcinai no problem, more are coming soon ;-)

PR #43, opened 2015-07-16
Application getAuthorizeUrl returns whole URL + PHPUnit tests + .travis.yml

mhlavac:
`XeroPHP\Application::getAuthorizeURL` if you pass argument to it then it will return whole authorize url.

``` php
echo $application->getAuthorizeURL('AABB');
// outputs https://api.xero.com/oauth/Authorize?oauth_token=AABB
```

I've also added `phpunit.xml.dist` and `.travis.yml`.

You can now run tests by simply writting `phpunit` inside project ;-).

PR #52, opened 2015-07-17
Query improvements and Query parameter handling

mhlavac:
I had to start this changeset from #43 in order to use tests to test functionallity properly

Added features:
* Query Parameters
* Query includeArchived
* Query now converts some php values to Xero queries automatically (DateTime, string, int, float and boolean)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Using composer:
 
 ```json
   "require": {
-  	"calcinai/xero-php": "1.2.*"
+    "calcinai/xero-php": "~1.2"
   }
 ```
 
@@ -67,9 +67,9 @@ Or create & populate it
 ```php
 $contact = new \XeroPHP\Models\Accounting\Contact($xero);
 $contact->setName('Test Contact')
-	->setFirstName('Test')
-	->setLastName('Contact')
-	->setEmailAddress('test@example.com');
+    ->setFirstName('Test')
+    ->setLastName('Contact')
+    ->setEmailAddress('test@example.com');
 ```
 
 Save it
@@ -91,7 +91,7 @@ Attachments
 ```php
 $attachments = $invoice->getAttachments();
 foreach($attachment as $attachment){
-	//Do something with them
+    //Do something with them
     file_put_contents($attachment->getFileName(), $attachment->getContent());
 }
 
@@ -103,3 +103,12 @@ $invoice->addAttachment($attachment);
 PDF - Models that support PDF export will inherit a ```->getPDF()``` method, which returns the raw content of the PDF.  Currently this is limited to Invoices and CreditNotes.
 
 Refer to the [examples](examples) for more complex usage and nested/related objects.
+
+Query parameters
+``` php
+$query = $xero->load('Accounting\\Contact');
+$this->query->where('Name == :name OR FirstName == :first_name');
+$this->query->setParameter('name', 'ACME');
+$this->query->setParameter('first_name', 'John');
+$contacts = $query->execute();
+```
--- a/src/XeroPHP/Remote/Query.php
+++ b/src/XeroPHP/Remote/Query.php
@@ -11,7 +11,9 @@ class Query {
     const ORDER_ASC  = 'ASC';
     const ORDER_DESC = 'DESC';
 
-    /** @var  \XeroPHP\Application */
+    /**
+     * @var Application
+     */
     private $app;
 
     private $from_class;
@@ -20,6 +22,16 @@ class Query {
     private $modifiedAfter;
     private $page;
     private $offset;
+
+    /**
+     * @var bool
+     */
+    private $includeArchived = false;
+
+    /**
+     * @var mixed[]
+     */
+    private $parameters = [];
 
     public function __construct(Application $app) {
         $this->app = $app;
@@ -42,22 +54,64 @@ class Query {
     }
 
     /**
+     * @param string $where
+     * @param mixed|null $value
      * @return $this
      */
-    public function where() {
-        $args = func_get_args();
-
-        if(func_num_args() === 2) {
-            $this->where[] = sprintf('%s=="%s"', $args[0], $args[1]);
+    public function where($where, $value = null) {
+        if ($where && $value) {
+            $parameterName = sha1($value);
+            $this->where[] = sprintf('%s==%s', $where, ":$parameterName");
+            $this->setParameter($parameterName, $value);
         } else {
-            $this->where[] = $args[0];
+            $this->where[] = $where;
         }
 
         return $this;
     }
 
+    /**
+     * @param string $where
+     * @param mixed|null $value
+     * @return $this
+     */
+    public function andWhere($where, $value = null) {
+        return $this->where($where, $value);
+    }
+
+    /**
+     * @return string
+     */
     public function getWhere() {
-        return implode(' AND ', $this->where);
+        return implode(' AND ', array_map(function($where) {
+            return "($where)";
+        }, $this->where));
+    }
+
+    /**
+     * @param string $key
+     * @param mixed $value
+     * @return $this
+     */
+    public function setParameter($key, $value)
+    {
+        $this->parameters[$key] = $value;
+
+        return $this;
+    }
+
+    /**
+     * @param mixed[] $values
+     * @return $this
+     */
+    public function setParameters($values)
+    {
+        $this->parameters = [];
+        foreach ($values as $key => $value) {
+            $this->setParameter($key, $value);
+        }
+
+        return $this;
     }
 
     /**
@@ -86,7 +140,7 @@ class Query {
     }
 
     /**
-     * @param int $page
+     * @param int|null $page
      * @return $this
      * @throws Exception
      */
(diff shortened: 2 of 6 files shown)


src/XeroPHP/Remote/Query.php
```diff
@@ -34,20 +50,65 @@ public function from($class) {
         return $this;
     }
 
-    public function where() {
-        $args = func_get_args();
```
calcinai: I think the reason I did this in the first place was so you could explicitly pass ```null```

mhlavac: @calcinai you can still pass null, but now you know what are you passing.

I would suggest changing the API to have `where` and `andWhere` methods.

`where` would override whatever `where` was used before and `andWhere` would add additional statements with and.

Doctrine QueryBuilder does it the same way and it's really convenient

mhlavac:
@calcinai I've rebased this PR to master and it can be merged now.

README.md
```diff
@@ -100,3 +100,12 @@ $invoice->addAttachment($attachment);
...
 
 Refer to the [examples](examples) for more complex usage and nested/related objects.
+
+Query parameters
+``` php
+$query = $xero->load('Accounting\\Contact');
+$this->query->where('Name == :name OR FirstName == :first_name');
+$this->query->setParameter('name', 'ACME');
+$this->query->setParameter('first_name', 'John');
+$contacts = $query->execute();
```
mhlavac: @calcinai this is example of parameters usage.

mhlavac:
@calcinai is this fine? Should I add more documentation or something? What do you think about it?

calcinai:
Sorry - I've been away from the office for the last couple of weeks and still not back! Only issue is that there are quite a few loose types and comparisons floating around in there. Meant to get back to you sooner with some suggestions!

mhlavac:
@calcinai I'll be glad for any suggestion. You can see tests to see how those things convert.

src/XeroPHP/Remote/Query.php
```diff
@@ -20,6 +25,16 @@ class Query {
     private $page;
     private $offset;
 
+    /**
+     * @var bool|null
+     */
+    private $includeArchived = null;
```
calcinai: Should this be ```false``` by default since it's a default flag?

src/XeroPHP/Remote/Query.php
```diff
@@ -105,16 +162,50 @@ public function offset($offset = 0) {
     }
 
     /**
+     * @param bool|null $includeArchived
```
calcinai: ```bool```/```null```? what's the difference between passing null or false?

src/XeroPHP/Remote/Query.php
mhlavac: @calcinai if it's null then it won't be added to query at all. If it's true or false then it's added to query as true or false.

mhlavac: @calcinai we set it to false and not add it to query if it is false, because by default includeArchived is not shown anyway.

mhlavac:
@calcinai thank you for code review. I've fixed issue with null.

src/XeroPHP/Remote/Query.php
```diff
@@ -135,22 +221,45 @@ public function execute() {
...
+    /**
+     * @param string $value
+     * @return string
+     */
+    private function convertPHPValueToXeroQuery($value)
+    {
+        if ($value instanceof \DateTime) {
+            return sprintf('DateTime(%s)', $value->format('Y,m,d'));
+        } else if (is_int($value) || is_float($value)) {
+            return $value;
```
calcinai: In this line, you're basically saying if it's an int or a float, return it.  Even though the function returns a 'string'  I think it'd be appropriate to cast it here.  It's a shame we can't share almost duplicate functionality [here](https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/Object.php#L227)

src/XeroPHP/Remote/Query.php
```diff
@@ -41,22 +55,64 @@ public function from($class) {
...
+     * @param mixed|null $value
      * @return $this
      */
-    public function where() {
-        $args = func_get_args();
-
-        if(func_num_args() === 2) {
-            $this->where[] = sprintf('%s=="%s"', $args[0], $args[1]);
+    public function where($where, $value = null) {
+        if ($where && $value) {
```
calcinai: What is this actually testing with ```$where``` and ```$value``` and what if you want to pass ```0``` as a condition?  Please use strict type comparisons everywhere for clarity.  (I don't even think you need to be testing $where here as it's a mandatory parameter).  A better test would be ```if($value !=== null)```

Also, it seems a little risky sha1'ing the value - what happens if two clauses have the same value?

src/XeroPHP/Remote/Query.php
```diff
@@ -105,16 +161,46 @@ public function offset($offset = 0) {
     }
 
     /**
+     * @param bool $includeArchived
+     * @return $this
+     */
+    public function includeArchived($includeArchived = true) {
```
calcinai: What's the point of this function accepting false as a parameter? Won't it just be left out when the params are built?  Another case for strict type checks for clarity on https://github.com/calcinai/xero-php/pull/52/files#diff-677c4a583ee02cfb477d433b7cbc2d55R224

mhlavac: @calcinai the reason for parameter here is that you can includeArchived by just using it, or you can disable the condition by calling includeArchived(false);

``` php
// Prepare your query in some method
$query->includeArchived(); // archived are going to be included

// In other method I get prepared query
if ($something === true) {
    $query ->includeArchived(false); // will not include archived anymore.
}
```

I don't need this functionality for myself. So we can remove it for now and make it argument less, but I think for anyone who would need this it might be a nice feature.

mhlavac:
@calcinai thanks for the review. I implemented suggestions from your comments + I've added 2 exceptions to check if parameter is unused or missing.

src/XeroPHP/Remote/Query.php
```diff
@@ -42,22 +60,63 @@ public function from($class) {
...
      */
-    public function where() {
-        $args = func_get_args();
-
-        if(func_num_args() === 2) {
-            $this->where[] = sprintf('%s=="%s"', $args[0], $args[1]);
+    public function where($where, $value = null) {
+        if ($value !== null) {
+            $parameterName = $this->addInternalParameter($value);
+            $this->where[] = sprintf('%s==%s', $where, ":$parameterName");
```
mhlavac: @calcinai I was also not sure with the sha1 for parameter code. I used `internalParameters` instead. Anything added via where with value argument will be added to internal parameter list and then translated to final value.

We have to do this, because otherwise value that starts with colon could get translated to parameter value.

tests/Remote/QueryTest.php
```diff
@@ -0,0 +1,157 @@
...
+        $this->assertArrayNotHasKey('includedArchived', $this->query->getRequest()->getParameters(), 'includeArchive should not be set in Request if set to false');
+    }
+
+    /**
+     * @expectedException \XeroPHP\Remote\Exception\QueryParameterNotSetException
+     */
+    public function testParameterNotSetThrowsException()
+    {
+        $this->query->where('test == :name');
+        $this->query->getRequest();
```
mhlavac: @calcinai exception is thrown, because `:name` parameter is missing

tests/Remote/QueryTest.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @expectedException \XeroPHP\Remote\Exception\QueryParameterNotUsedException
+     */
+    public function testParameterNotUsedThrowsException()
+    {
+        $this->query->where('test == :name');
+        $this->query->setParameter('name', 'something');
+        $this->query->setParameter('email', 'test@example.com');
+        $this->query->getRequest();
```
mhlavac: @calcinai exception is thrown because email parameter is not used anywhere

src/XeroPHP/Remote/Query.php
calcinai: That makes sense.  I guess it was just tripping me up why it would sometimes be null.

src/XeroPHP/Remote/Query.php
calcinai: That seems now to be overcomplicated!  I was just meaning to use something that'd be unique instead of that sha1.

calcinai:
Would you still like this merged?  If so can you please rebase it again?

mhlavac:
@calcinai I'll get to this today. I'll rebase it and look into it if I can do it in less complicated way. We already use the query parameters and I find them really useful. I think it's a nice addition to the library

mhlavac:
@calcinai I am sorry that this go so much delayed... I hope to have a look soon

calcinai:
How are you going with this?

mhlavac:
@calcinai sorry, there is always something. Does it make sense to finish this for `1.x`, or should I wait till `2.x`?

calcinai:
Maybe we just wait for 2.x, it's not far off.

PR #54, opened 2015-07-21
Add missing UpdatedDateUTC property

coop182:
Added a bit of code to the generator to find the UpdatedDateUTC property if it is available.


coop182:
I have not got a commit for the updated models yet... someone at Xero has been having a great time playing with the documentation and have broken the parsing of Tax Rates and Payments.

I have some ideas on how to get around the Tax Rates stuff basically hard coding the $section to TaxType when $section = TaxRate and $subsection = TaxType... but its pretty nasty and I don't like the idea of special cases....

The Payment model is breaking because someone has listed Payment on the types page twice...

calcinai:
Yep - I was doing some work with this last night - ran into exactly the same issue.

I think I've almost got it sorted.

calcinai:
This isn't necessary anymore I don't think.

PR #56, opened 2015-07-21
Add a request callback and a method of adding other callbacks

gerrywastaken:
I've added a user callback for when we are making requests to Xero

This is useful for users in the following instances:

* allowing counting of the number of requests being made to Xero (important due to the very low 1000 requests per day limit)
* logging requests being made
* automated sanity checks to make sure their code is resulting in the desired requests

I'm sure others will find other uses for this that I haven't thought of. 

Note: I know I could have made this a lot more flexible, such as allowing multiple callbacks per event, having the return affect weather the call to xero is actually executed, etc. But instead I have opted for keeping the changes small and easy to understand. I haven't added a test for this yet but will.

### DIFF
--- a/examples/config.php
+++ b/examples/config.php
@@ -35,5 +35,17 @@ $xero_base_config = array(
         //CURLOPT_SSLKEYPASSWD    => '1234',
         //CURLOPT_SSLKEY          => 'certs/entrust-private-RQ3.pem'
 
+    ),
+
+    // Functions which will be invoked at certain points during execution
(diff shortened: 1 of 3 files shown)


calcinai:
Have you had a look at #23? Same sort of idea. 

My only comments on this are that I don't think the callback registering belongs in the Xero confit array..

gerrywastaken:
Ha! Ahh damn. I hadn't seen that. Yup that solution would be better if everybody is on board with it.

PR #59, opened 2015-07-27
Add __isset magic function to Remote\Object

steveh:
Some properties are not sent from the Xero API, for example `Contact.AccountNumber`.

If you try and call `getAccountNumber()`you will receive a notice that the index doesn't exist on `_data`. Overloading this function means you can test it first.

The alternative option would be for the getters to test if the property exists and return `NULL` if it doesn't.


calcinai:
Looks good to me. Thanks

PR #60, opened 2015-07-27
Isset, getter & setter tests for Object + Object ArrayAcces

mhlavac:
1. Adds tests for #59 + getters and setters in general.
2. Adds ArrayAccess to object, to allow accessing Object values like Array `$contact['Name']`

@calcinai @steveh I found out that newly added `__isset` is case sensitive, while `__get` & `__set` is not. The reason behind is that methods in PHP are not case sensitive, but arrays are.

### DIFF
--- a/src/XeroPHP/Remote/Object.php
+++ b/src/XeroPHP/Remote/Object.php
@@ -10,7 +10,7 @@ use XeroPHP\Helpers;
  * Class Object
  * @package XeroPHP\Remote
  */
-abstract class Object implements ObjectInterface, \JsonSerializable {
+abstract class Object implements ObjectInterface, \JsonSerializable, \ArrayAccess {
 
     /**
      * Keys for the meta properties array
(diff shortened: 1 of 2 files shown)


steveh:
That's a good point. Perhaps the keys of _data could be converted to lowercase?

calcinai:
I remember having issues with the Xero API somewhere early on with case sensitivity, so I tries to keep everything CS as in the docs.. I can't imagine the same model will have different props with the same name in a different case so should be ok.

mhlavac:
@steveh @calcinai so do you want to do make get and set case sensitive as well?

Currently you can do this

``` php
echo $contact->firstName;
echo $contact->FirstName;
echo $contact->FiRsTnAmE;
```

and they are all valid. If we change this behaviour now, then we should put it in 2.x branch because it's BC break.

I can make ArrayAccess case sensitive only, because it's new feature that can't be used by anyone at the moment.

calcinai:
I'll merge this and we can have a closer look in 2.0

PR #69, opened 2015-08-08
Other apis support

ronanq:
I found a couple of problems that prevented the payroll and files APIs from working correctly.


calcinai:
Thanks for that @roranq - clearly no one has used anything other than the Accounting API up until now!

PR #72, opened 2015-08-11
Fix private example

mhlavac:
Public key is not required for `Private Application`, but the other certificate is.


calcinai:
Thanks Martin.

Issue #76, opened 2015-08-14
Models\Accounting\Attachment must implement abstract method isPageable()


drawmyattention:
The class ```XeroPHP\Models\Accounting\Attachment``` extends ```Object``` and therefore must implement the abstract method ```isPageable()```.

Adding:

    static function isPageable()
    {
        return false;
    }

should fix this issue.

calcinai:
Closed by #78

drawmyattention:
Can you release an update with this patch included? As it stands, I can't do a composer update because my autoloader throws an exception because of the missing method implementation.

calcinai:
Sorry - done!

PR #77, opened 2015-08-14
Add isPageable method

drawmyattention:
Fixes issue #76

### DIFF
--- a/src/XeroPHP/Models/Accounting/Attachment.php
+++ b/src/XeroPHP/Models/Accounting/Attachment.php
@@ -204,7 +204,7 @@ class Attachment extends Object {
     /**
      * @return mixed
      */
-    static function isPageable() {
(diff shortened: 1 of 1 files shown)


src/XeroPHP/Models/Accounting/Attachment.php
```diff
@@ -201,4 +201,11 @@ public function setLocalHandle($local_handle) {
         $this->local_handle = $local_handle;
     }
 
+    /**
+     * @return mixed
+     */
+    static function isPageable() {
```
calcinai: Visibility?

drawmyattention: Good spot. Rather lazily, I copied the function from the abstract class and didn't check. Will resubmit a new PR.

PR #78, opened 2015-08-17
Attachment fix

drawmyattention:
Fixes issue #76


calcinai:
Awesome, thanks.

calcinai:
Also minor, but I wonder why the doc is @return mixed? Whoops

PR #83, opened 2015-08-21
Application loadByGUID handles payroll entities

digiwise:
Tried to use
```PHP
$xero->loadByGUID('PayrollAU\\Employee', $guid)
```
But it didn't use the Payroll API.  Hope this fixes it.  This is my first ever git pull request so hope it is right!


calcinai:
Nice, thanks.  This is actually the other half of #69!

Issue #80, opened 2015-08-18
Putting dates into the correct timezone


judgej:
This issue has a long stream of my ramblings, trying to get to the bottom of how Xero accepts, stores and presents dates:

https://github.com/calcinai/xero-php/issues/79

The result of my testing is: dates are provided by the API as strings exactly as typed by the user in the Xero admin pages. It makes no difference what the organisation timezone is set to, either when the records were created, or now - the date is as typed. I have not explored dates supplied in the PUT or POST API schemas, as I am just dealing with GET at the moment.

So when reading these dates, they need to be interpreted in the timezone of the organisation. Those timezones in Xero are UCT plus or minus an offset, and do not appear to have a DST. I don't know if non-DST timezones are supported by PHP - I suspect not (but would love to be corrected).

I think at the very least, when configuring the application, you should be able to provide the timezone to be used when interpreting these dates, e.g. "Europe/London".

The workaround for now is to set the default system timezone, which is a bit of a hack IMO (dates and time parsing should never rely on the local timezone when dealing with remote data - it needs to be more explicit).

Thoughts?

FLamparski:
In my previous job, I scraped the surface of this strange intersection of timezones and general Xero API silliness - I can't remember exactly what we did but I think we interpreted the UTC dates as UTC, and it worked out for us.

Also note that converting between the Windows Registry timezone names and PHP's timezones (based on the IANA database) is... _fun_.

judgej:
Oh yes, those timezone conversions really do look like an awful lot of fun :-(

I've had confirmation from a Xero dev that any date field in the XML API results with a name that ends in "UTC" should be treated as UTC.

In my own tests, all other dates (even though Xero presents them as datetimes with a 00:00:00 time) are returned by the API *as typed by the end user*. So they have no real timezone (they are just dates after all) so you need to be careful what timezone you put them into if you want to do any arithmetic on them. This seems to hold true even when dates are entered between DST dates, and regardless of what Xero timezone is set for the organisation. Something niggles the back of my mind that this could come back to bite us, but I've not put my finger on it yet.

For my project now, we are defining the IANA timezone for an organisatino *outside* of Xero. That seems to be the safest option - we don't care what timezone is used to display times to the end Xero user, and we know exactly what dates they have typed into the screens (as that is what the API gives us) so we parse them according to the timezone we define.

Issue #79, opened 2015-08-17
Timezone for UTC times should be UTC.


judgej:
Fetching a transaction I see this in the transaction:

                        [UpdatedDateUTC] => DateTime Object
                                (
                                    [date] => 2015-08-17 12:34:28.260000
                                    [timezone_type] => 3
                                    [timezone] => Europe/London
                                )

I'm guessing the timezone comes from the account timezone, but this timestamp from Xero is special - it is a UTC time and should not be presented in the timezone of the account.

So far as I can see the "date" element is correct if interpreted as UTC, but wrong if interpreted as a "Europe/London" timezone time.

judgej:
The timezone might even be being picked up locally. I've not investigated where it is coming from yet, just noting that what is presented by the API is wrong - the time is correct, but only as a UTC time.

judgej:
By setting the local default timezone, I get this:

    date_default_timezone_set('UTC');

    [UpdatedDateUTC] => DateTime Object
                (
                    [date] => 2015-08-17 12:34:28.260000
                    [timezone_type] => 3
                    [timezone] => UTC
                )

Notice the timezone in the timestamp returned by the API is now UTC, but the time string is not. By setting my local time, this means the time the API library is giving me, is *completely different* - it has jumped backwards by an hour.

It is important to keep the timezone of the remote Xero account when interpreting times/date values that don't have one explicitly in the API data. In this case the API field is *specifically* giving us a UTC time, so that meeds to be handled by the API library.

Dates are going to be a whole other walk in the park, since PHP does not have a date type. I'll come back to those later.

judgej:
Here is the raw API data for the last update time:

    <UpdatedDateUTC>2015-08-17T12:34:28.23</UpdatedDateUTC>

Only the name gives us any indication of the timezone. *However* that may be a Xero bug - other timestamps in the API *do* give us the timezone. For example the current time on the organisation:

    <DateTimeUTC>2015-08-17T17:19:07.5133435Z</DateTimeUTC>

Note the "Z" on the end - that's UTC, that is. See http://stackoverflow.com/questions/8405087/what-is-this-date-format-2011-08-12t201746-384z

Raw data for dates on the same API fetch of the payment:

        <Date>2015-08-10T00:00:00</Date>
        <DueDate>2015-08-27T00:00:00</DueDate>

[...]

PR #85, opened 2015-08-31
UTC support

calcinai:
Fixes? #79 and #80


judgej:
Looks fun - I'll have a play with this. So I need to be aware that some return values that used to be strings are now `\DateTime` types?

calcinai:
Yeah, sorry that's an unfortunate side effect of making it actually pick up those columns.

judgej:
NP (edited previous comment to add more :-)

PR #107, opened 2015-10-14
Add support for setting bank account number into Accounting/Account model

wisoot:


wisoot:
This code should be generated instead

Issue #93, opened 2015-09-16
Filter function to search for records. Lack of examples.


mediabeastnz:
Filter function to search for records. ~~Currently only GUID is searchable~~

Maybe this issue can be for the lack of examples around filtering?
For example how would do multiple elements with AND as doesn't the where function only take two variables?

FLamparski:
You can drop a ->where() between load and execute to filter records. Check
the Xero API docs for more info. Also note that not all searching and
filtering may be possible in the API that is possible in the UI.

coop182:
@mediabeastnz the `where()` function can take one or two arguments

So for multiple where conditions with AND you can call it in two ways:

```php
$xero->load('Accounting\\Invoices')->where('Status','AUTHORISED')->where('Type', 'ACCREC')->execute();
```
```php
$xero->load('Accounting\\Invoices')->where('Status=="AUTHORISED" AND Type=="ACCREC"')->execute();
```
If you want to do OR you have to use the single argument method as multiple `where()` calls are automatically put together with an AND as you can see [here](https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/Query.php#L59).

mediabeastnz:
Thank you for clarifying this!

PR #109, opened 2015-10-27
Added coop182's examples for filtering to readme

phbaudin:
see https://github.com/calcinai/xero-php/issues/93

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,11 +60,17 @@ foreach ($contacts as $contact) {
 
 Search for objects meeting certain criteria
 ```php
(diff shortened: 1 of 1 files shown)


calcinai:
Do you think it's worth using the class constants in here?

phbaudin:
Indeed, I didn't know there were some. If the instructions don't promote best practices then what will!

I'll change that tomorrow.

On 27 oct. 2015, at 19:16, Michael Calcinai <notifications@github.com <mailto:notifications@github.com> > wrote:

Do you think it's worth using the class constants in here?

coop182:
Yeah that was me being lazy haha... should have used the constants!

PR #111, opened 2015-11-19
Handling bad requests with messages for invalid fields

M1ke:
I was finding it difficult to debug when receiving the following Exception:

> [XeroPHP\Remote\Exception\BadRequestException] A validation exception occurred (An error occurred in Xero. Check the API Status page http://status.developer.xero.com for current service status. Contact the API support team at api@xero.com for more assistance)

occurring and not being able to debug, as the response was not available in scope. This was a bit frustrating doing TDD through Codeception, as that test framework doesn't work well with Xdebug.

I made a quick change to parse the ValidationErrors into a set of messages that get appended to the root error.

It could likely have presentation or code style improvements and with some feedback I'd be happy to make them!

If there's an alternative way of doing the same thing please let me know too.

### DIFF
--- a/src/XeroPHP/Remote/Response.php
+++ b/src/XeroPHP/Remote/Response.php
@@ -73,6 +73,7 @@ class Response {
                 //This catches actual app errors
                 if(isset($this->root_error)) {
                     $message = sprintf('%s (%s)', $this->root_error['message'], implode(', ', $this->element_errors));
+                    $message .= $this->parseBadRequest();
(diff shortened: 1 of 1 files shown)


calcinai:
Looks ok to me - I know Xero can be particularly unhelpful when it comes to the actual error messages.

M1ke:
Thanks for the merge. This is a fantastic library, very useful to quickly develop Xero applications, so I'll try and pitch in on more issues in future!

Issue #118, opened 2015-12-03
DateTimeInterface instead of DateTime


M1ke:
Having received fatal errors when trying to pass a `DateTimeImmutable` into a set method I'm wondering if we should alter all typehints to allow the interface rather than the specific object.

I'll do the work; from what I can see a simple replace of:

* `(\DateTime` => `(\DateTimeInterface`
* `@property \DateTime` => `@property \DateTimeInterface`
* `@return \DateTime` => `@return \DateTimeInterface`

Would work, however you might have more of an insight. Let me know if that would seem sensible.

calcinai:
I don't see any reason that this wouldn't work.  There's a bit of stuff around changing timezones for UTC which might be the only issue?

Have you tested it?

M1ke:
I've submitted a PR for this:

https://github.com/calcinai/xero-php/pull/119

calcinai:
Hey, did you use the generator to make the changes to the models?  (Just before I merge this in)

M1ke:
Nope just made text changes. What's the process? Can try and see what's
different

calcinai:
Oh, sorry I don't see this! you can just run ```./generator/generate``` from the root of the project.

calcinai:
Have you had a chance to run the generator?

M1ke:
I've now run the generator and pushed up to my master branch:

https://github.com/sturents/xero-php/commit/675fe35647e040d80d355b24c1db46f1bbb11d7b

Has this worked as expected?

PR #119, opened 2015-12-04
Change DateTime to DateTimeInterface

M1ke:
The ability to choose to use the older `\DateTime` object or the newer `\DateTimeImmutable`, or even any other date library which doesn't extend `\DateTime` but uses `\DateTimeInterface` improves the flexibility of this library.

This PR alters:

* `\DateTime` typehints to `\DateTimeInterface`
* `@return` phpDoc comments
* `@property` phpDoc comments
* `@param` phpDoc comments
* `@var` phpDoc comments
* Object generator strings from "\\DateTime" to "\\DateTimeInterface"

All tests are passing. No new tests have been added.


M1ke:
Based on discussion on:

https://github.com/calcinai/xero-php/issues/118

src/XeroPHP/Remote/Response.php
```diff
@@ -120,10 +120,10 @@ private function parseBadRequest(){
 			$field_errors = [];
 			foreach ($this->elements as $n => $element){
 				if (isset($element['ValidationErrors'])){
-					$errors[] = $element['ValidationErrors'][0]['Message'];
```
M1ke: This change doesn't follow the rest of the PR, but it's a minor thing I noticed whilst testing which was a bug in my previous accepted PR, so thought I may as well fix it.

calcinai:
Did you see my comment re: the generator? The PR currently contains changes to generated code,

PR #127, opened 2016-01-29
Public.php Demo & Query Logic

sketchthat:
Updated logic in `examples/public.php`

- `getOAuthSession` compares the expires session to the current time. Variable wasn't a timestamp it was an integer of how many seconds until the token expired. Logic updated to save session as `time() + seconds`.
- Array key doesn't exist in example, changed `expires` to `oauth_expires_in`.

Updated logic in  `src/XeroPHP/Remote/Query.php`

- `where` expected query to be `==` but if a boolean is passed for queries like `Contact.IsCustomer = "TRUE"` only request one `=` operator. [Could use `->where('IsCustomer="TRUE"')` if this request isn't merged].

### DIFF
--- a/examples/public.php
+++ b/examples/public.php
@@ -79,7 +79,7 @@ print_r($xero->load('Accounting\\Organisation')->execute());
 function setOAuthSession($token, $secret, $expires = null){
     // expires sends back an int
     if($expires !== null) {
-        $expires = time() + $expires;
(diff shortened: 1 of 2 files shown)


calcinai:
OAuth logic needed some work, thanks.  What strange behaviour with those bools.

PR #141, opened 2016-02-17
Added getWarnings method to Invoice object to return any Xero warnings

netmit:
Added quick method of returning Xero warnings from Invoice object when creating an invoice without having to create a request object.


calcinai:
I think there are some circumstances where this won't be reliable.  The warnings shouldn't come from he OAuth client, so I think it's after to use a longhand approach like this instead:

```php
$response = $xero->save($invoice);
print_r($response->getElementWarnings());
```

Or similar.

Issue #142, opened 2016-02-17
Save invoice action does not populate line item data


radonjic:
Hi,

I have notice that when Invoice record is save, InvoiceID is populate from response, but Line Item Id is not populate. 
Can you add small fix by adding optional parameter to save method in Application class and to paste it in fromStringArray method(see attachment).

Thanks in advance,

Ivan 
<img width="1440" alt="xero_php" src="https://cloud.githubusercontent.com/assets/7442790/13111596/fb2c715a-d586-11e5-9892-3a6c32911bef.png">

calcinai:
Will merge when this is rebased.

PR #143, opened 2016-02-17
Add optional param to save method. Issue #142

radonjic:
Hi,

I have notice that when Invoice record is save, InvoiceID is populate from response, but Line Item Id is not populate. 
Can you add small fix by adding optional parameter to save method in Application class and to paste it in fromStringArray method. So that user can decide if they want to force object reload from response.

Thanks,

Ivan


calcinai:
Sorry, can you please rebase this and I'll merge it.

radonjic:
Hi,

Sorry for late response I was out off the office for a while. I have rebase my changes, it is a easy fix, I hope you will merge it soon.

Thanks in advance, 

Ivan

PR #146, opened 2016-03-04
Update README.md with Rate Limit Example

iveoles:


calcinai:
Do you think this should be a more generic example, e.g. not reliant on Laravel?  After reading it it actually seems like it's opening a can of worms trying to suggest how best to handle it.  It's going to differ in different applications.

The example you've given would lend itself to a batch job, in which case could it not just use a static variable inside a function to maintain state?  I hate to admit to the fact that there are some people that will just copy+paste that sort of thing and end up stalling a user's page load for a second if it needs to wait for the rate limit timeout.  There's also a potential race condition.

I think due to the sluggishness of the Xero API, we should encourage that any API transactions happen via a cron/beanstalk/mq or similar.

At the least, maybe we just give an example catching the ```RateLimitExceededException``` as (believe it or not) a lot of people still don't seem to completely understand try/catching different exceptions types in php.  I've lost count of how many times I've needed to refer someone to the php docs on this project alone!

As I said, can of worms. What to do?

iveoles:
Yeah this is used in crons and queues, so the pause isn't a big deal. I see
what you mean though, it could give bad performance if copy and pasted.

Perhaps it is best to leave it alone.

PR #150, opened 2016-03-16
Made InventoryAssetAccountCode not required

andy1547:


calcinai:
Hey, unfortunately I can't merge this as it's part of the generated code.  It'll just get overwritten next time the generator is run.  If the Xero documentation is correct, I'll update the parser to red it properly.

PR #152, opened 2016-03-23
Alter `DateTime` to `DateTimeInterface` and add better error messages

M1ke:
This PR does two distinct things:

* Alter all instances of `DateTime` as typehints, docblocks or other references to `DateTimeInterface` so that `DateTimeImmutable` or other public libraries adding date functions can be used.
* In `src/XeroPHP/Remote/Response.php` adds a parser to extract specific key/value errors so that the exception message contains more clear detail of what went wrong.

The generator has been run after code modification, and the branch has been merged with the latest master. All phpunit tests pass.


calcinai:
Hey, cheers for updating this.  I'm not sure if it's just the order that you've committed files, but just checking that you didn't manually change any of those files to DateTimeInterface?  If you did, they'll just get overwritten by the generator.

Also, I see you've added missing filed at b99401f - did they just come from upstream?

M1ke:
I originally changed all files as I didn't know about the generator. I then committed that. Then I ran the generator, which will have overwritten a number of those files, so it's not a problem.

Yes I also merged in upstream, as this branch has been a while since creation.

calcinai:
Awesome thanks, just wanted to make sure so your changes didn't get blown away.  

(can't see whether your changes were made by the generator or manually)

PR #156, opened 2016-04-11
isPropertyDirty dose not exists should be isDirty

tiran133:
In Your application you call a method isPropertyDirty
https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Application.php#L276

I tried to allocate a overpayment to an invoice. Since this Property is save Directly it tries to call this method.

But it should just be isDirty($porperty)

Alex

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -273,7 +273,7 @@ abstract class Application {
      */
     private function savePropertiesDirectly(Remote\Object $object){
         foreach($object::getProperties() as $property_name => $meta){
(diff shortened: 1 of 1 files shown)


calcinai:
Thanks for spotting this one.  Bust have been a BC introduced at some point.

tiran133:
Hey Thank you. 

Is it possible you tag this version so i will get the fix with an composer update?

Thanks
Alex

calcinai:
Hey, it's actually tagged as 1.3.0, since there are some other updated that might affect users of the 1.2 release.

I'll update the readme now to reflect this.

tiran133:
Hey thanks i just saw that. I will try this version see if it breaks anything in my app.

PR #172, opened 2016-06-15
Changing initial array variable name in case no errors are found

iveoles:
It turns out the initial variable had the wrong name. I must have been too tired when I looked at it yesterday.


calcinai:
Cool, thanks for that

PR #174, opened 2016-06-21
More unique nonce generation, date query params, reports

specialtactics:
I've had a lot of errors from the xero api saying my nonce has already been used when making many requests in a row, and I've tried a number of approaches at better nonce generation, and this seems to be working quite well.

### DIFF
--- a/src/XeroPHP/Models/Accounting/Report/BalanceSheet.php
+++ b/src/XeroPHP/Models/Accounting/Report/BalanceSheet.php
@@ -1,0 +1,261 @@
+<?php
+namespace XeroPHP\Models\Accounting\Report;
+
+use XeroPHP\Remote;
+
+class BalanceSheet extends Remote\Object
+{
+
+    /**
+     * Xero identifier
+     *
+     * @property string ReportID
+     */
+
+    /**
+     * Report Name
+     *
+     * @property string ReportName
+     */
+
+    /**
+     * Report Type
+     *
+     * @property string ReportType
+     */
+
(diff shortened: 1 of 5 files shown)


specialtactics:
Also added to & from query parameters.

I've also added functionality to use Reports. They seem to be fairly generic, so I've just included code for the 2 I've used so far, I will be looking further into it in the coming weeks.

calcinai:
Thanks for your work on this.  The noonce generation was following Xero's library, I will test to make sure it had no other repercussions. 

The report thing was something that's been missing for a while, I hadn't got around to parsing it out of the docs!

One comment re: the toDate and fromDate, how would you feel about them being date time objects passed, then formatted when the query is executed - that would be more in line with the conventions elsewhere.

specialtactics:
Hi there

I do prefer the use of DateTime, though in this case because Xero only took date, I figured it may be confusing in the sense that people might assume you can specify a time where it actually would get discarded. So I'm happy to defer to you on this one, I don't really mind either way.

specialtactics:
Just wondering what the status of this is. Would you still prefer to be using the DateTime objects following on from that discussion, and are you looking for me to implement that?

calcinai:
Sorry about not getting back to you. I've been away for the last week. If you didn't mind changing them to DateTimes, that would be great. Otherwise it's fine to merge.

specialtactics:
No worries, I've updated the PR, switched to DateTime (looking pretty clean), and added another date query param.

PR #175, opened 2016-07-06
Corrected Accounting\Invoices to Accounting\Invoice

olimortimer:
Fixes 'Class does not exist [\XeroPHP\Models\Accounting\Invoices' error


calcinai:
Ha, thanks - I noticed that the other day and forgot to fix it.

PR #178, opened 2016-07-26
Add orWhere() and andWhere() methods to Query class

WebSpanner:
This PR adds an orWhere() method to allow OR WHERE statements in queries which was not previously available. The PR also adds an andWhere() method which is an alias for where() (which would already produce an AND statement when chained on to another where() statement). The purpose of the andWhere() method is to optionally improve readability and clarity of fluent queries in the code.

The logic of the where() method has been refactored into a new method called addWhere() which when chained will add either an AND or an OR statement depending on the $operator argument which is passed in as the first argument. The where() method remains and behaves the same to keep the API consistent and backward compatible.

Queries can now be added like this.

    $this->xero
        ->load('Accounting\\Contact')
        ->where('Name', 'Steve')
        ->orWhere('Name', 'Nancy')
        ->execute();

### DIFF
--- a/src/XeroPHP/Remote/Query.php
+++ b/src/XeroPHP/Remote/Query.php
@@ -42,14 +42,48 @@ class Query {
 
         return $this;
     }
-
+    
     /**
+     * Adds a WHERE statment to the query. Can also be used to chain an AND WHERE statement to
+     * a query.
+     *
      * @return $this
      */
     public function where() {
-        $args = func_get_args();
+        return $this->addWhere('AND', func_get_args());
+    }
+    
+    /**
+     * Chains an OR WHERE statement on to the query
+     *
+     * @return $this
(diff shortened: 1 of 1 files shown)


calcinai:
Looks ok to me.  In 2.0, I have refactored this to be quite similar - you can do things like 

```php
$this->xero
    ->load('Accounting\\Contact')
    ->where('Name', 'Steve')
        ->or('Name', 'Michael')
    ->orWhere('Name', 'Nancy')
        ->and('IsSupplier', true)
```

That way lets you go two levels deep with nested and/or conditions.  Thanks for your work on this though - I'm sure it'll be useful until 2.0 is stable.

WebSpanner:
Awesome thanks! I didn't know about those changes in 2.0. Looking forward to it.

PR #184, opened 2016-08-17
Payroll PayItems fix

iamthefox:
Hello,

I submitted similar request for the payslip endpoint last night. 

Basically API returns results in as a single node. 

Might be worth looking in more abstract solution for this problem. 

Thank you,
Alex

### DIFF
--- a/src/XeroPHP/Remote/Response.php
+++ b/src/XeroPHP/Remote/Response.php
@@ -239,8 +239,8 @@ class Response {
                     $this->root_error['message'] = (string) $root_child;
                     break;
                 case 'Payslip':
-                    // payslip API always returns Payslip as a single node rather than array of nodes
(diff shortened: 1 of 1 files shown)


calcinai:
Have you used other payroll classes?  Do you know if this behaviour is actually documented anywhere?

iamthefox:
yup, I used PayRuns, PayItems, Payslip (in singular) and Employee. They have API preview tool but you will need to register an account, to see what output is going to look like

iamthefox:
by the way, only these 2 had issues with it

calcinai:
What a pain in the.

iamthefox:
yeah, unfortunately it is not the most consistent API :(

calcinai:
Well thanks for adding these workarounds.  It's thinks like this that are slowing down 2.0!

Issue #185, opened 2016-08-19
Contact - TrackingCategory properties incorrect


Sixfifty:
I'm encountering an issue when loading contacts that have a SalesTrackingCategory. It is not loading the properties on the TrackingCategory correctly, it seems that Xero might have changed the properties they are returning.

This issue is currently causing Invoicing to fail for me. I have a small fix which appears to work.

I added these two items to the array in the `getProperties()` function in `Accounting\TrackingCategory` and it has solved the issue for me:

`'TrackingCategoryName' => array (false, self::PROPERTY_TYPE_STRING, null, false, false),`
`'TrackingOptionName' => array (false, self::PROPERTY_TYPE_STRING, null, false, false),`

Is this the best way to fix this issue? If it is then how can I create a pull request?

This is my first bug report, let me know if I can improve on my reporting as well.

calcinai:
There is very little in the documentation regarding the use of tracking categories.  Can you please attach some example XML?

Sixfifty:
Here is a text file containing the full XML, I have removed personal details.

[SampleContact.txt](https://github.com/calcinai/xero-php/files/427023/SampleContact.txt)

The key part is this:
```
      <SalesTrackingCategories>
        <SalesTrackingCategory>
          <TrackingCategoryName>Clients</TrackingCategoryName>
          <TrackingOptionName>Saunderson</TrackingOptionName>
        </SalesTrackingCategory>
      </SalesTrackingCategories>
```

Sixfifty:
Ah, I see what you mean by scraping their API documentation now.

Their API does not match what they are sending back to me:

https://developer.xero.com/documentation/api/tracking-categories/

I'll drop Xero a line with this as well. I have a bug report open with them regarding an issue that was caused by this.

Sixfifty:
Xero have now updated their documentation on this page to show the change:

https://developer.xero.com/documentation/api/contacts/#title14

The don't appear to have updated the page for tracking categories. Worth me chasing them about that?

calcinai:
Great! If you can that'd be good - it's in their interests to have accurate docs!

Sixfifty:
Message sent, let's see if they update it.

Sixfifty:
Fantastic! I'll give that a try when I get back onto the project in a few days.

Thanks for helping!

calcinai:
All I had to do was type `generator/generate` and commit the code.  You and Xero did all the work!

00dani:
[...]

PR #186, opened 2016-08-19
Added TrackingCategoryName and TrackingOptionName properties.

Sixfifty:
This is in reference to Issue #185


calcinai:
Hey, since this is generated code, I can't merge this.

Sixfifty:
Fair play, I hadn't quite grasped that it was a live generation.

calcinai:
Funny story - I actually only wrote the scraper to build the initial models, to try and save a bit of time in the beginning.  It stuck around because Xero often change things in the API without actually telling people (like removing/renaming properties), so this lets it put `deprecated

PR #199, opened 2016-10-13
Update and clear invoice line item functionality

john-main-croud:
Amended addLineItem function so that if there is a matching line item ID it will update rather than add.
Added function to clear all line items.

### DIFF
--- a/src/XeroPHP/Models/Accounting/Invoice.php
+++ b/src/XeroPHP/Models/Accounting/Invoice.php
@@ -385,7 +385,27 @@ class Invoice extends Remote\Object
         if(!isset($this->_data['LineItems'])){
             $this->_data['LineItems'] = new Remote\Collection();
         }
+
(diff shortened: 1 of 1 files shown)


calcinai:
Hey, unfortunately I can't merge this as it's in generated code.  What might be more appropriate is if we adjust the way collections add/remove items and go from there?

john-main-croud:
Yeah sure thing - all I really need is a mechanism to update rather than exclusively add, so that I am able to handle amended LineItems.
Might be worth having a specific remove action too - I just suggested the clear all as a  quick fix.

calcinai:
Hey, so I've just had a look at the collection class - there's already the ability to clear via the `->removeAll()` method e.g.

```php
$invoice->getLineItems()->removeAll()
```

I'll make a method now for updating an object if it already exists.

Issue #46, opened 2015-07-17
PSR-2 Coding Standard


mhlavac:
Is anyone opened to use PSR-2 coding standard?

http://www.php-fig.org/psr/psr-2/

calcinai:
Definitely not a bad idea to pick a standard. I previously used PHPStorms auto-format I've the whole project. 

I might have to get used to opening function brackets on new lines!

mhlavac:
@calcinai this can help a lot http://cs.sensiolabs.org/ ;-)

gerrywastaken:
Indeed I hate the psr standards as they use votes instead of solid reasoning behind their decisions, but unfortunately it's been adopted by the PHP community in general so there is really no escaping it.

@calcinai I hate the opening bracket thing too.

I've used the cs fixer tool mentioned by @calcinai and it's actually very good so long as you haven't got inconsistent indenting anywhere.

mhlavac:
@calcinai @gerrywastaken PSR standards are not so bad. There are a lot of discussions going on before voting starts, it's same as PHP RFC a long discussions and then it finally goes down to votes, because there is always someone who doesn't like the final solution and discussions could go on forever.

I was also against spaces and opening brackets on new line, but I got used to it and have no problem with it anymore. I try to respect standard syntax for different languages, because it's easier to contribute that way.

gerrywastaken:
@mhlavac I disagree that it's not so bad. But I do agree that it's a good idea to follow it.

calcinai:
That sounds like a decision..

gerrywastaken:
@calcinai Well it's your repo. I can't make the decision. But if it's a yes then I think this issue can be closed and we can all start coding to it (right @mhlavac?). :)

mhlavac:
@calcinai @gerrywastaken right. I can open PR later this week with reformatted code to PSR2. Good thing is that cs fixer can be used on generated methods as well ;-)

calcinai:
I think it'd be better to fix the generator templates to conform, otherwise, every time we generate it'll put a whole lot of rubbish in the git history.

calcinai:
But yes, let's do it for the rest of the code.  Why not!

PR #200, opened 2016-10-18
Code cleanup

Synchro:
I've done some small non-functional changes in this PR. Mainly conversion to short array syntax and correction of some phpdocs. Some of this will help with #46. I did not do any conversions to the models, but I amended the generator template to use short array syntax so that they will be done next time that's run.

In `save()` I made it return early if the Object is not dirty, helps reduce unneeded indenting.

There were a couple of undeclared properties in the generator (`sub_models` and `supports_page`) which I added declarations for.

These changes help spot errors in IDEs like PHPStorm.

### DIFF
--- a/examples/config-test.php
+++ b/examples/config-test.php
@@ -10,23 +10,23 @@ use XeroPHP\Application\PartnerApplication;
 
 require '../vendor/autoload.php';
 
-$config = array(
-    'oauth' => array(
+$config = [
+    'oauth' => [
         'callback'    => 'http://localhost/',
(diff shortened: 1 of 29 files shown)


calcinai:
Cool, I'm all for changes like this - the long array syntax was a hangover from supporting PHP 5.3

calcinai:
Also, good point int he same method - I have no idea how it came to be like that!

Synchro:
Thanks, I'll see what else I can find!

PR #201, opened 2016-10-18
PSR-2 code cleanup

Synchro:
Lots of changes for PSR-2 in #46, mostly layout things, plus some phpdoc tweaks. Still needs work, but this should be the bulk of it! Does @mhlavac have anything to add?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Using composer:
 
 ```json
   "require": {
-  	"calcinai/xero-php": "1.3.*"
+    "calcinai/xero-php": "1.3.*"
   }
 ```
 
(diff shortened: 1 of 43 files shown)


calcinai:
Haha, a fair few changes there.  It looks some stray newlines between arguments are still scattered throughout.

Are you aware of a standard for inline doc comments to hint vars?  e.g.
```
/** @var Object $object */
```

I looks a little strange them being 3 lines in the middle of code just for the sake of an IDE.

Synchro:
I once did this for HTMLPurifier - 16,000 changes in 450 files in one PR! 😱 

Not sure what you mean by newlines between args?

I'm not sure about that - it's not PHPStorm that's complaining about those but phpcs. The default phpcs profile is extremely fussy about phpdocs - can't be single-line, must contain a comment, tags need to be in specific order, must have some blank lines etc. Them being bigger doesn't bother me, but I'll look at finding a better profile.

calcinai:
I agree with all of those points in actual comments, just the type hints for the IDE seem a little overkill.  No big deal really though.

calcinai:
Something small I've noticed regarding the multiline IDE hints is that 
```
/** @noinspection PhpUndefinedVariableInspection */
```
 works, but
```
/** 
 * @noinspection PhpUndefinedVariableInspection
 */
```
Seems to be ignored by PHPStorm.  I know those are edge cases but still useful sometimes (when the IDE has no idea what's going on!)

PR #202, opened 2016-10-19
More PSR-2 changes!

### LINKED ISSUE #46: PSR-2 Coding Standard

Synchro:
Hello again... I ran all the code (minus Models) through [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) with `level=psr2` and it made a bunch of changes, nearly all layout related. 2 notable changes - it added 'public' prefixes where they were missing, and changed the capitalisation of the `XeroPHP\Tests\Application` namespace to `XeroPHP\tests\Application` so that it matches the `test` directory name. I don't think that has any negative consequences.

There are still a load of phpdocs missing that might help auto-completion if we get around to them at some point, but for now I think you can probably close #46.


calcinai:
Nice.  I've also realised that those whitespace things I was mentioning was the wrapping on the page (duh). 

Thanks for tidying this up - tests pass, so it can't have interrupted that process.

PR #203, opened 2016-10-20
Add config files

Synchro:
This PR includes a `.php_cs` config file for php-cs-fixer making it easier to run the fixer on the project source. It's configured to exclude the models folder as fixes for that should be done in the generator.

I also added a `.gitattributes` file which is used to suppress the inclusion of non-essential files when the library is pulled in from composer or downloaded as a zip file - for example it skips the examples, generator, and test folders which are not needed at runtime. This is considered a good practice by [the League](http://thephpleague.com).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+* text=auto
+
+/docs export-ignore
+/tests export-ignore
+/examples export-ignore
+/generator export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.php_cs export-ignore
+/.travis.yml export-ignore
+/changelog.md export-ignore
+/phpunit.xml.dist export-ignore
+/README.md export-ignore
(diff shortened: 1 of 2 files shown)


calcinai:
Looks good to me!  Thanks for your work on this

Synchro:
Happy to help! It's all because I've been having a total nightmare trying to work with the Xero supplied OAuthSimple code. I'm giving up on that and switching everything to your lib, which is going to mean a major rewrite of our sync plugin for [1CRM](https://www.syniah.com), but hopefully it will be much better that way! Having all the static analysis / autocomplete working nicely in PHPStorm helps keep the quality up.

I really do hate OAuth with a vengeance, so anything to make it simpler is a good thing!

calcinai:
OAuth itself aside, what baffles me is how Xero can actually endorse their own implementation.

PR #207, opened 2016-11-16
Add reference to 3rd-party Symfony bundle

foaly-nr1:


calcinai:
Are there any other bungles/providers/facades that should be mentioned here?

foaly-nr1:
Not sure. I'm the maintainer of the XeroBundle – I wrote it because I couldn't find an existing one when I looked back then.

PR #215, opened 2016-12-22
Removed entrust ssl certificates

SidneyAllen:
In 2017, Xero is deprecating the entrust ssl certificate requirement
for partner apps.  You can find more information at Xero's Dev Center - http://developer.xero.com/documentation/auth-and-limits/entrust-certificate-deprecation/

### DIFF
--- a/src/XeroPHP/Application/PartnerApplication.php
+++ b/src/XeroPHP/Application/PartnerApplication.php
@@ -6,10 +6,5 @@ use XeroPHP\Application;
 
 class PartnerApplication extends Application
 {
-    protected static $_type_config_defaults = [
-        'xero' => [
-            'site'     => 'https://api.xero.com',
-            'base_url' => 'https://api.xero.com',
-        ]
-    ];
+    protected static $_type_config_defaults = [];
 }
--- a/src/XeroPHP/Application/PartnerApplication.php
+++ b/src/XeroPHP/Application/PartnerApplication.php
@@ -8,8 +8,8 @@ class PartnerApplication extends Application
 {
     protected static $_type_config_defaults = [
         'xero' => [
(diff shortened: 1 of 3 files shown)


src/XeroPHP/Application/PartnerApplication.php
```diff
@@ -8,8 +8,8 @@ class PartnerApplication extends Application
 {
     protected static $_type_config_defaults = [
         'xero' => [
-            'site'     => 'https://api-partner.network.xero.com',
-            'base_url' => 'https://api-partner.network.xero.com',
+            'site'     => 'https://api.xero.com',
+            'base_url' => 'https://api.xero.com',
```
calcinai: Thanks for update @SidneyAllen - I was aware of it but have just implemented it int he `2.0` branch as it's nearing being ready.  These settings actually only override the defaults, so if they're no different to the other APIs, the keys can be removed entirely.

SidneyAllen: Ok - so if I understand correctly - you've already made the update and will ignore my pull request.  Any idea when you'll release 2.0?   All new App Partners are not getting entrust certificates so we're actively updating libraries to support these developers.

An email about this change to all existing App partners is scheduled for Jan 13th

calcinai: I'm happy to merge most of it, just these two lines are redundant, since they're defined [here](https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Application.php#L17)

SidneyAllen:
Ok - I've removed the config default urls from PartnerApplication.php

calcinai:
Nice, thanks for updating that.

PR #218, opened 2017-01-15
Make inventory asset account code not compulsory

stralytic:
Xero API doesn't require it for saving new Items.


calcinai:
Hey, can you please try to update the model generator, not the models directly as these changes will just be overwritten. 

Also, I'm not sure why you've changed the package name...

PR #219, opened 2017-01-15
Accounting/Item->inventoryAssetAccountCode is optional

stralytic:
Make generator realise that 'on a tracked inventory item' means that $mandatory is false.


calcinai:
Thanks for your work on this.  I've just noticed something else (my fault this time) that the generator isn't respecting deprecated functions.  I'm going to have to fix this before this can be merged.

calcinai:
@stralytic If possible, can you please update the models to not remove those functions then I can merge? I'm trying to establish a baseline as the new models are built from the OpenAPI spec

hajima03:
How to add address in contact model ?

PR #236, opened 2017-03-23
getNonce(): Only use mt_rand to generate a random string

### LINKED ISSUE #235: getNonce() assumes microtime() will have a decimal part
The getNonce() method in XeroPHP/Remote/OAuth/Client.php assumes that microtime will return a decimal part, and explodes on the decimal point:
https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/OAuth/Client.php#L232

However, it is possible that microtime will land exactly on the second, and no decimal part will be returned.

Josh-G:
Fixes #235

### DIFF
--- a/src/XeroPHP/Remote/OAuth/Client.php
+++ b/src/XeroPHP/Remote/OAuth/Client.php
@@ -227,11 +227,9 @@ class Client
      */
     private function getNonce($length = 20)
     {
-        // Add more uniqueness to the nonce
-        $parts = explode('.', microtime(true));
-        $nonce = base_convert($parts[1], 10, 36);
+        $nonce = "";
 
-        for ($i = 0; $i < $length - strlen($nonce); $i++) {
+        while (strlen($nonce) < $length) {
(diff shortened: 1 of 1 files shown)


Josh-G:
Is just having mt_rand not random enough?

https://github.com/calcinai/xero-php/commit/71db90c3a0f962048786c6203ab8a4c0a027d4b5#diff-48c7813bb7d60d193f619de8f20cecd4

calcinai:
I never actually encountered this issue - as you saw, it was changed last year because I assume the noonce was colliding..

Josh-G:
Would it not be less random replacing 4 random characters from mt_rand with the current time?

Was Xero saying you were colliding? Isn't the probability of that incredibly rare? Sorry, I'm a little confused here :D

Josh-G:
I think this change makes more sense. It's currently an issue for us as we elevate warnings into exceptions and this warning occurs 10-20 times a day despite it's rarity.

PR #237, opened 2017-03-29
Fix misplaced comment in Application->save()

direvus:
The comment "In this case it's new" was shown above the branch where it's *not* new.


calcinai:
Thanks for picking this up

PR #239, opened 2017-03-30
Fix incorrect example of loadByGUID in README

direvus:
The example previously appeared to pass the GUID in an array: `[GUID]`, whereas loadByGUID actually expects a single GUID as string.


calcinai:
Haha, sorry this was wasn't so clear - this was just a placeholder showing that that was the parameter for the `guid`

It's definitely more clear to show it as var.

PR #245, opened 2017-04-10
fix PayItem request body, fix request for employee on PayrollAU

rizaljzulfikar:


calcinai:
This has since been resolved

PR #249, opened 2017-04-20
Add methods to get and set config options

Josh-G:
It can be useful for us to check and update config options on the fly without having to re-instantiate the Application. We do this at the moment to try and workaround Xero timeouts by selectively increasing the curl timeouts when performing an important operation.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -126,7 +126,7 @@ abstract class Application
         $this->config = array_replace_recursive(
             self::$_config_defaults,
             static::$_type_config_defaults,
-            $user_config
+            $config
         );
 
         return $this->config;
@@ -272,7 +272,7 @@ abstract class Application
     public function saveAll($objects)
     {
         $objects = array_values($objects);
-        
+
         //Just get one type to compare with, doesn't matter which.
         $current_object = $objects[0];
         /**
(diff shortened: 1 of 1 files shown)


Josh-G:
@calcinai: Any interest in getting this merged?

calcinai:
Whoops sorry, will review this afternoon

src/XeroPHP/Application.php
```diff
@@ -108,6 +104,48 @@ public function getConfig($key)
...
+     * @param mixed $option
+     * @param mixed $value
+     * @return array
+     * @throws Exception
+     */
+    public function setConfigOption($config, $option, $value) {
+        if (!isset($this->config[$config])) {
+            throw new Exception("Invalid configuration key [$key]");
+        }
+        $this->config[$config][$option] = $value;
```
calcinai: Does this make an assumption that the config will always be 2D?

Josh-G: Thanks for reviewing.
It assumes it will be like the $_config_defaults array, the first key is the group name (e.g, xero) and the second key is the option name. E.g., I would be calling setConfigOption('curl', 'CURLOPT_USERAGENT', 'Josh-G'); to set the useragent. If you'd like the behaviour changed at all, please let me know and I will amend the pull req

Josh-G:
@calcinai: sorry to prod again, any concerns in getting this merged?

Issue #252, opened 2017-05-03
Payments not returned in CreditNotes


thekeogh:
`Payments` aren't being returned with `CreditNotes`.

Adding the following to the `Account\CreditNote` `getProperties()` fixes it:

```
'Payments' => [false, self::PROPERTY_TYPE_OBJECT, 'Accounting\\Payment', true, false],
```

Can we get this added? I'm unable to make a PR currently. thanks

PR #258, opened 2017-05-22
Added Payments property to CreditNote

thekeogh:
Although the Xero docs do not document a `Payments` property exists on `CreditNotes`, it does, as credit notes get payments associated to them when you apply refunds to them.

Without this change you cannot see what refund payments have been made against the CreditNote. Fix for issue #252


calcinai:
This looks ok to me, although does it also need a getter?

thekeogh:
I may very well do, I never needed a getter, but it should deffo have one. I can update my PR tomorrow with it.

thekeogh:
I have added the getter, my change passes all tests, but looks like Travis died on a composer install. Doesn't look like an updated dependency failure, seems to be a Composer error of `Content-Length mismatch` of which I have seen before in old versions of composer, a composer self-update fixed that. Can you try re-running the build, probably just a gremlin.

My change passes though, so that can be merged in to the master.

calcinai:
Managed to run test again successfully.  Cheers for that

PR #261, opened 2017-05-30
Add Line objects for Employee PayTemplates

wbercx:
PayTemplates consist of:
* `EarningsLines`
* `DeductionLines`
* `SuperLines`
* `ReimbursementLines`
* `LeaveLines`

Some of those already have existing variations within `Payslips`, which are currently used in PayTemplates, but the fields are different. As such, I have created dedicated objects for each of the line types for use within `PayTemplates`.

Any issues, let me know :)


calcinai:
Looks good to me, thanks for this.  Hopefully we can get the model definition more robust in future

PR #268, opened 2017-07-18
Remove non-minimum config settings

timacdonald:
When I was digging into this, I was a bit confused as to what I was reading on Xero and in this config file, then I realised this was not actually apart of the minimum config for a private application. Thought removing it might help people in the future coming to this.


calcinai:
Yeah, you're probably correct. There are actually few cases where this is required anyway

Issue #246, opened 2017-04-10
Get the Invoice Url (OnlineInvoice)


cezzre:
I've been searching a while for this method – which seems to be absent.
I eventually managed to get it working.
I think it is worth to add this method to the Invoice object.

Ref: https://developer.xero.com/documentation/api/invoices#onlineinvoice

```php
private function getInvoiceUrl(\XeroPHP\Models\Accounting\Invoice $invoice)
{
    $url = new \XeroPHP\Remote\URL($this->conn, 'Invoices/' . $invoice->getGUID() . '/OnlineInvoice');
    $request = new \XeroPHP\Remote\Request($this->conn, $url);
    $request->send();

    return $request->getResponse()->getElements()[0]['OnlineInvoiceUrl'];
}
```

`$this->conn`  – which is a property of the class I've created to _talk_ with Xero – is the instance of `PrivateApplication` (in my case).

timacdonald:
@cezzre that pull request to add this feature has been pulled in, so once a new version is out you should be right to close this issue 👍

PR #269, opened 2017-07-18
Ability to get the online invoice url [#246]

timacdonald:
Been using this today and needed this functionality. I adapted this from @cezzre issue #246 

This would, in a perfect world, maybe not just be some methods but a class itself as it's kinda a entity in the Xero API world, but I think for now this will suffice. Can always refactor it out later.


calcinai:
Looks fine to me, I'll need to remember that this exists when regenerating the models!

PR #270, opened 2017-07-18
Fix spelling of Australian Constants

timacdonald:
I've found several constants that actually reference the chemical element `AUSTRALIUM` 🤓 instead of the intended `AUSTRALIAN`

I've renamed these, but kept the originals - best to drop in a major version release - but added a deprecated comment.


calcinai:
Hey, thanks for this - this must've been how it was spelt in the docs!

timacdonald:
haha - no worries at all 👍

PR #274, opened 2017-07-25
Throw exception when accessing Public app in Private context

circle314:
Fixed an issue where an Exception was not thrown when trying to access a Public application using a Private application context. The response code in this case is 400, but the error is stored in $this->oauth_response['oauth_problem_advice'], rather than $this->root_error['message']


calcinai:
This is covered in a later PR

PR #277, opened 2017-08-01
Allow FQNs for models

timacdonald:
I think it would be nice to allow the use of [FQNs using the class constant](http://php.net/manual/en/language.oop5.constants.php#example-184)

```php
$xero->load(Contact::class)
```

I personally prefer this style over:

```php
$xero->load('Accounting\Contact')
```

I've added the ability to use the FQN, but also fallback to appending the config namespace. I added some tests for this as well.

**Note** I removed the `$class[0] !== '\\'` condition as I figure it's going to fail whether that check is there or not, and it just clutters the method (I personally feel) - but let me know if you want me to add it back in.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -156,16 +156,29 @@ abstract class Application
      */
     public function validateModelClass($class)
     {
-        $config = $this->getConfig('xero');
-
-        if ($class[0] !== '\\') {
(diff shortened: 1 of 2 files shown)


timacdonald:
I just added another test to cover the removal of the `$class[0] !== '\\'` check, just to be safe.

calcinai:
This is great - the thought had crossed my mind a few times, thanks.

PR #279, opened 2017-08-01
Remove static config variables in preference for basic config variable.

timacdonald:
I've removed the static config options and consolidated everything into the single `$config` property. A couple of reasons.

### Duplication.

Once an Application is instantiated, it has the instance `$config` and the classes static `$_config_defaults`.

### Children Class type variables in base class

Having `$_config_defaults` in the base `Application` class and then a `$_type_config_defaults` also in the base class just for the children classes felt a bit off, as the base class shouldn't really be concerned or care how the children classes are implemented.

All this made the `setConfig` method overly complicated keeping track of `static::`, `self::`, `$config` etc.

### Standardisation

The right way to set config options in the children class was not a clear path for users. There is the `__construct($config)` method, but also the `$_type` config. I figure the idea is the `$_type` config was really for internal use only, but I also figure we can just use the `__construct($config)` internally to help keep things simple.

### Possible Breaking Change [Possibly actually a Bugfix]

With the current code base the following test fails because whenever `setConfig` is called, the default configs are used to override current configs, i.e. when we call the second `setConfig` method, the `site` value is overridden by the static variables default value.

```php
$site = 'http://my.custom.site/';
$app = $this->instance();
$app->setConfig(['xero' => ['site' => $site]]);
$app->setConfig(['xero' => ['core_version' => '9.9']]);

$this->assertEquals(
    $site,
    $app->getConfigOption('xero', 'site')
);
```

With this PR, this test passes (as I personally would expect it to) and I've also included the test in the PR.

---

All in all, I think this moves to a cleaner class having a single `$config` variable to deal with.

---
Side note: @calcinai I hope you don't mind me doing some PR's. I like to help out on the project as I've got a bit of spare time and thought I could help test the code and make a few adjustments I thought might help. I've used it on my latest project, so would like to say thanks by helping out. Just let me know if these are unwanted or too much.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -11,7 +11,10 @@ use XeroPHP\Remote\URL;
 
 abstract class Application
 {
-    protected static $_config_defaults = [
+    /**
+     * @var array
+     */
+    protected $config = [
         'xero'  => [
             'site'            => 'https://api.xero.com',
             'base_url'        => 'https://api.xero.com',
@@ -42,19 +45,10 @@ abstract class Application
     ];
 
     /**
-     * @var array
-     */
-    protected $config;
-
-    /**
      * @var Client
      */
     protected $oauth_client;
 
-    /**
-     * @var array
-     */
-    protected static $_type_config_defaults = [];
 
     /**
      * @param array $user_config
@@ -123,11 +117,7 @@ abstract class Application
      * @return array
      */
     public function setConfig($config) {
-        $this->config = array_replace_recursive(
-            self::$_config_defaults,
-            static::$_type_config_defaults,
-            $config
-        );
+        $this->config = array_replace_recursive($this->config, $config);
 
         return $this->config;
     }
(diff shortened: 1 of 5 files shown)


timacdonald:
I guess one thing to consider is if end devs would have been using the removed variables in their extended classes. If you think they have been, this is probably going to have to be a breaking change release. i.e. x.x.0 version

calcinai:
Hey, no issue at all. As you can probably imagine, I've got a list a mile long with things I'd like to do here. I'm still trying to get 2.0 out, would you be interested in helping on that?

timacdonald:
Yea, for sure. I'm not 100% across the code base, but I would be down for helping out. I can't guarantee any specific time each week, but in my spare time would be more than happy to help out and push things along if I can!

timacdonald:
@calcinai I might re-work this PR to make it a little less breaky and perhaps deprecate instead. I'll try and work on it today and send it through and you can decide if its the path ya wanna go down with it.

calcinai:
Awesome, cheers

timacdonald:
I think this is probably to much of a *possible* breaking change. Are you actively dev'ing a version 2?

I think this would be better for that. The idea here was to consolidate the config into one store.

PR #280, opened 2017-08-02
Moved authorize url query string logic to Client

timacdonald:
So in the `Application` there was logic to prepare the url retrieved from the `Client` via the `getAuthorizeURL()` method. I figure that the URL logic is kinda best to be taken care of by the `Client`, and that the `Application` just has a `getAuthorizeURL()` method as a (proxy) convenience.

Also, if someone was to call:

```php
$application->getOAuthClient()->getAuthorizeURL('MY_TOKEN');
```

it would not have had the query string attached as previously the `Application` was doing that, with this PR however, the query string would now be attached correctly.

Could probably extract a URL Helper class to deal with the url stuff I moved, but its a bit cleaner now anyway.

+ Tests!

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -81,14 +81,7 @@ abstract class Application
      */
     public function getAuthorizeURL($oauth_token = null)
     {
-        $authorize_url = $this->oauth_client->getAuthorizeURL();
-
-        if ($oauth_token !== null) {
-            $operator = parse_url($authorize_url, PHP_URL_QUERY) !== null ? '&' : '?';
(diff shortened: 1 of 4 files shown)


calcinai:
Nice

timacdonald:
I'll fix up these conflicts tonight.

timacdonald:
That's been fixed.

timacdonald:
@calcinai I went a bit quiet on this cause I threw so much at ya at once haha.

What are your thoughts on this one? I don't see it being a breaking change.

calcinai:
Let's do it

timacdonald:
Damn! that was quick haha

Issue #259, opened 2017-05-24
Removing LineItems from an Invoice


mrchimp:
As far as I can see there's no way to remove a LineItem from an existing Invoice. According to the [Documentation](https://developer.xero.com/documentation/api/invoices#modify-invoice) you can remove a LineItem by excluding it with an update request.

Assuming I'm not missing something and there isn't already a way to do this, I'm happy to submit a pull request. The only way I can see to do it at the moment is to create a new Invoice, apply all the existing data to i except for the relevant LineItem and then save that but that seems like a roundabout way to do things.

mnwalker:
Easier than that, just need to modify the $this->_data['LineItems'] and then save.
Already got a working version so will submit a PR

calcinai:
Is it too counterintuitive to do something like:
```
$invoice->getLineItems()->removeAt($x)
```
?

calcinai:
This is now supported via the collection.

PR #282, opened 2017-08-03
Added 2 new methods to remove all or individual line items.

mnwalker:


mnwalker:
#259

calcinai:
Hey, do you think this is a better approach than just calling `->removeAll()` on the collection itself?  Only for the sake of consistency across models...

mnwalker:
Didn't spot that was available, will be easier yes.

PR #288, opened 2017-08-05
Refactor exceptions

timacdonald:
Hey,

Just removed a bit logic that isn't needed in these exceptions. The message and code are still able to be overridden by passing the the values to the constructor when throwing an exception, this is just a nicer way to set defaults.

I also added the blank comment `//` to empty classes. This is totally just a personal preference thing. If you don't like it, happy to remove it and stick to whatever standard you prefer.

I didn't add any tests, as we would just be testing PHP behaviour and it didn't seem needed.

Again, tell me to slow down if this is unwanted stuff or whatever. Just enjoying working through a code base and seeing what I can help out on.

### DIFF
--- a/src/XeroPHP/Exception.php
+++ b/src/XeroPHP/Exception.php
@@ -4,4 +4,5 @@ namespace XeroPHP;
 
 class Exception extends \Exception
 {
+    //
 }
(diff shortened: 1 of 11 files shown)


calcinai:
Who knows what was going on here!

Issue #271, opened 2017-07-20
[Discussion] Refactor the Application out of the models


timacdonald:
Hey @calcinai,

So this is probably a massive task - but I would be happy to help.

I get why the application is in the models. It allows for a ORM style `$invoice->save()` which I love. However there is an issue when you serialize the models, it contains the application AND any responses tied to the application.

This was an issue I had to work around - where I was adding jobs the a database queue and needed to pass the invoice instance between jobs. It was serialized and saved no worries - until I hit the job to download the PDF of the invoice as the application IN the model still contained all the PDF data and was causing encoding issues, not to mentioned it just adds bloat to the serialized model.

Or is the better solution here to simply remove the application when serializing the models. Considering thats the quickest fix...now I've written this, I probably think that would be the best solution haha.

calcinai:
Yeah, can the base model just implement Serializable and JsonSerializable, then just overload it to return the data only?

timacdonald:
Yea, if we just `$this->_application = null` in those to remove that? Would that cause any issues you can imagine?

We're you thinking more of just removing the request responses from the application but keeping the application in the model during serialization? That way when you restore the model you don't have to add the application again?

calcinai:
The main issue here is actually just the fact the oauth client is setting that `request` property. I think this can be easily removed and just passed into the relevant functions..

timacdonald:
hmmm - what if we just a `removeApplication` method. Then everything is a little more explicit. You don't have to do any magic to make sure things are deserialized properly etc. And it doesn't have an BC issues.

timacdonald:
I mean I could just pass in a fresh instance of the app when I serialize, but I think just having the ability to remove it would be good - because it will still have to serialize that object - which I'd prefer not to be storing if possible - because it would have credentials etc in it as well.

PR #294, opened 2017-08-10
Removed member property for request in OAuth client

calcinai:
This PR removes the `$request` property from the OAuth client. This stops objects containing all of the previous request data.

This possibly addresses #271 too.

### DIFF
--- a/src/XeroPHP/Remote/OAuth/Client.php
+++ b/src/XeroPHP/Remote/OAuth/Client.php
@@ -30,11 +30,6 @@ class Client
 
     private $config;
 
-    /**
-     * @var Request $request
-     */
-    private $request;
-
     /*
      * "Cached" parameters - will change between signings.
      */
@@ -62,10 +57,8 @@ class Client
      */
     public function sign(Request $request)
     {
-        $this->request = $request;
-
         $oauth_params = $this->getOAuthParams();
-        $oauth_params['oauth_signature'] = $this->getSignature();
+        $oauth_params['oauth_signature'] = $this->getSignature($request);
 
         //put it where it needs to go in the request
         switch ($this->config['signature_location']) {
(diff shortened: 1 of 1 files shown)


calcinai:
What do you think about this @timacdonald

timacdonald:
I do think this is good and a better way to go about things, rather then setting the request on the client.

The only issue I'd say is that it is changing the public API on `getSBS()` to `getSBS(Request $request)` so if you are doing Semantic Versioning with the project, it might be best targeted at the `2.0.0` release.

calcinai:
Yes, you're right.  That should really be private - it's of zero use anywhere else.  Probably wouldn't even work..

timacdonald:
Also, if the code 'feels' better because the client seems more like a utility to sign the request rather than something that owns the request. So passing it in feels better than having it saved to the object.

timacdonald:
One thing on my list of things (yes...I started an ideas list hehe) is to review all the visibilities through, see if we can't clean up the public API

calcinai:
Yeah exactly what I was thinking. Not sure what I was thinking a few years back when I wrote it - the whole point was to make something better than [this](https://github.com/XeroAPI/XeroOAuth-PHP/blob/master/lib/OAuthSimple.php)

timacdonald:
Oh....I know! I love that Xero actually promote your project over their own in the docs! Man, I think the same thing when I look back at things I wrote 6 months ago.

timacdonald:
This is an awesome project and I didn't really have any problems using it, it was dreamy. That serialize thing was my only issue, which I just sorted by passing the invoice ID between jobs instead of the invoice object.

timacdonald:
I personally reckon if you got this thing tested out and perhaps an automated docs system and few other tweaks, you should hit up Xero and say - ours is fully tested, documented, modern PHP. Make this the official one and ditch theirs.

calcinai:
On a semi-related note, I went to the 'XD17' talk the other day, and one of the main points they made was "DO NOT WRITE YOUR OWN SDK" 

Ha.

timacdonald:
Check it, your at the top haha: https://developer.xero.com/documentation/libraries/php

Oh interesting, was their suggestion let the community? I wrote a [wrapper around the Kumulos API](https://github.com/timacdonald/kumulos-api) and they loved it and then wanted to take over the repo...I'm just like - just merge into this one...it's not like it's anything special anyway! Uh!

Put a license on it...never heard from them again haha

PR #297, opened 2017-08-14
Add `getAPIStem()` to `saveAll()`

wbercx:
Just tried to push multiple Timesheets to Xero at once when I realised that `saveAll()` was trying to submit the data to the Core API rather than the Payroll API.

Unlike the `save()` method, `saveAll()` does not currently get the API stem for the objects you try to save. Since the API is not explicitly passed into the URL object when it is being created, it defaults to `self::API_CORE`.

My change includes a call to `$type::getAPIStem()` and passes the result into the URL object (just like `save()` does) so your data ends up in the right place.


calcinai:
Hey, thanks for this.

PR #298, opened 2017-08-14
Refactor PDF trait

timacdonald:
Pretty straightforward refactor of this trait. Just makes it a little cleaner (I think).

Because we are newing up in the methods I'm not sure of anyway to write unit tests for this trait yet, but I've added it to my list to work out and implement perhaps in a future major version. I've manually tested and it seems all sweet, but go over it if you are keen on the PR.

- made property access reply on abstract methods - that way we don't need to document what `$this` is as we don't care, all we care is it has the methods implemented.
- added `getApplication` to the remote object to satisfy the abstract method.
- methods docs
- extracted 2 smaller, more focused methods for building the `URL` and `Request`. I've called these `buildXYZ` so you know what they are doing i.e. they aren't accessing an already existent object, they are building a new one. Just a convention I've been using recently that I think I like...tell me if you hate it and would prefer something else.
- rearranged the `use` statements longest to shortest. This is totally a personal preference - if you prefer alphabetical or some other conversion let me know and I'll fix it up.

### DIFF
--- a/src/XeroPHP/Remote/Object.php
+++ b/src/XeroPHP/Remote/Object.php
@@ -85,6 +85,16 @@ abstract class Object implements ObjectInterface, \JsonSerializable, \ArrayAcces
     }
 
     /**
+     * Get the application.
+     *
+     * @return \XeroPHP\Application
+     */
+    protected function getApplication()
+    {
+        return $this->_application;
+    }
+
+    /**
      * If there have been any properties changed since load
      *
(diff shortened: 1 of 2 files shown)


timacdonald:
Okay, so apparently older (<7.0) versions of PHP don't like abstract static methods in traits. I've just commented that out for now and made a note that this needs to be looked in a bigger refactor.

PR #304, opened 2017-09-05
Update Xero docs link for pagination

jessedc:
The pagination/rate limiting link I've updated in this PR was returning a 404. I believe I've found the correct updated version.


calcinai:
Cheers for that!

PR #308, opened 2017-09-13
Fixed invalid XML creation for purchases and sales details

DrRoach:
When adding purchase and sales details as array elements the `arrayToXML` method in `Helpers.php` creates the following invalid XML:

```XML
<PurchaseDetails>
  <PurchaseDetail>
    <COGSAccountCode>300</COGSAccountCode>
  </PurchaseDetail>
</PurchaseDetails>
```

Setting both the purchase details and sales details as proposed in the pull request fixes this issue. This may however cause compatibility issues if adding multiple purchase or sales details but I don't think this is possible so shouldn't be an issue.


alexbowers:
@calcinai Can this be tagged and released?

Issue #315, opened 2017-10-08
Blocking: getNonce() causes random break.


edegaudenzi:
XeroPHP\Remote\OAuth::getNonce() randomly breaks the whole module due to the following bug:

microtime(true) returns the decimal part of the number (basically, the fractions of a second).

The problem is that used as is, it is very common to jump into the pitfall where the second is a pure integer number, no decimal part, causing the crash in the following explode() function that instead correctly expects something to split.

Examples:
Bugged microtime()
```
1507482512.97
1507482512.98
1507482512.99
1507482513
1507482513.01
1507482513.02
```

What we need instead:
```
1507482602.9999780654907226562500
1507482602.9999880790710449218750
1507482602.9999980926513671875000
1507482603.0000081062316894531250
1507482603.0000178813934326171875
1507482603.0000278949737548828125
```

Code to reproduce the bug (watch at the apache2 error.log):
```
<?php
for ($i=0; $i < 100000; $i++) { 
$microt = microtime(true);
echo explode('.', $microt)[1].'<br>';
sleep(0.1);
}
```

PR #316, opened 2017-10-08
#315 Fixed the bug introduced by the microtime() randomly missing decimal part.

edegaudenzi:


calcinai:
This piece of code has been changed so many times!

PR #328, opened 2017-12-14
Fix 'Undefined index: message'

aziev:
`$this->root_error` sometimes equals to empty array, so we need to check this case


aziev:
The problem was occurred because I was trying to make requests to public app instead of private. So in this case the variable `$this->root_error` was an empty array without both message and code. But the message "Bad request" will make more sense then "Undefined index: message".

However, I agree with you, that message and code should to be more specific.

If you can share with me thoughts about which code and message we should to use in this case it'll be very helpful. Then I'll be able to make new commit

Issue #326, opened 2017-11-09
PHP 7.2 support


stralytic:
Use of the keyword "object" is not allowed in any case in PHP 7.2, which prevents the use of XeroPHP with this version of PHP.

`Compile Error: Cannot use 'XeroPHP\Remote\Object' as class name as it is reserved`

circle314:
Good catch @stralytic , I imagine the circumstances around finding this were unfortunate. I would have encountered it myself but for the fact that our codebase won't immediately benefit from 7.2, and I planned on moving to 7.3 instead once that's out.

The comments on the class mention the following:

> todo - at 2.x, move this into the root of the project and refer to it as a model

@calcinai how far off is 2.x? Given this is a showstopper for anyone on PHP7.2 (and they're already working on PHP7.3) could this change be implemented in a v1.7 as a stopgap?

calcinai:
I *think* so, but it might be prudent to keep a class called `Object` in there extending whatever it is renamed to unless there is some type hinting somewhere on others' code.  Thoughts?

I had a chat to Xero a couple of days ago about moving forward - it looks like they're wanting to allocate a bit of response to this project, so watch this space. We went through a rough plan of where things are headed.

calcinai:
What was I thinking calling it `Object` anyway! It's hardly descriptive.

circle314:
Leaving it in there is fine, only so long as your composer.json file requires PHP <=7.1.*, though I would think that a safety measure rather than a future ready requirement. I've had the same problems with code I've written, sudden unexpected reserved word Colossians that is. It's gut wrenching :(

ilijastuden:
Now that PHP 7.2 is stable, this issue is causing brave souls that upgraded. What is the more descriptive name that you wanted to use? Maybe we can contribute a fix.

calcinai:
I think I'd thought to call it `Model` based on comments in the code! If you could submit a PR I'd happily review and merge it.

nigel-binaryio:
Thanks mattmcardle, appreciate your work here

ilijastuden:
Hi @calcinai 

I went to the mountains and brought only iPad, so I could not prepare and send a fix. When do you expect to have @nigel-binaryio's MR merged?

graphem:
Any news on this? Active Collab self hosted is broken on PHP 7.2 because of this. Thanks for all your work!!

calcinai:
Will test this afternoon and merge! Sorry for the delay

PR #331, opened 2018-01-15
Change XeroPHP\Remote\Object to  XeroPHP\Remote\Model

mattmcardle:
Fixes https://github.com/calcinai/xero-php/issues/326

### DIFF
--- a/generator/objects/Model.php
+++ b/generator/objects/Model.php
@@ -1,7 +1,7 @@
 <?php
 
 use XeroPHP\Helpers;
-use XeroPHP\Remote\Object;
+use XeroPHP\Remote\Model;
 use XeroPHP\Remote\Request;
 
 class Model implements ParsedObjectInterface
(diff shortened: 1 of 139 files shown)


calcinai:
This looks good, do you think it's worth adding an autoloader class `Object` extending `Model` for backward compatibility?

mattmcardle:
Added an empty abstract Object class that extends Model for backwards compatibility, is this what you meant @calcinai? I was a bit confused when you said "autoloader" class.

calcinai:
To be honest, I am also a bit confused by that comment.

But yes, that's what I meant, if you think it makes sense?

mattmcardle:
Yeah, I think its a good idea as it will prevent forcing people to make changes to references to  any references to `Object` they might have in their own code if they upgrade this package and are on < 7.2.

mattmcardle:
Just thinking, perhaps I should add a comment into the Object class to inform people it is only there for backwards compatibility? We could also take the opportunity to advise them to use the new Model class if they run into issues after upgrading to PHP 7.2 in the same comment, what do you think?

calcinai:
Absolutely - just adding the `@deprecated` tag to the class will inform most IDEs

mattmcardle:
Done, can you merge and do a release when you get a chance? I'm currently pointing to my fork :-1:

PR #334, opened 2018-01-26
Handle Oauth errors returned as bad requests

davidwindell:
One of our users had accidentally setup a public app instead of private.

Xero was returning the error ` Public applications must use the HMAC-SHA1 signature method` as a bad request instead of unauthorised where Oauth errors are usually caught. 

This was returning a blank error so this PR corrects that.

### DIFF
--- a/src/XeroPHP/Remote/Response.php
+++ b/src/XeroPHP/Remote/Response.php
@@ -116,7 +116,7 @@ class Response
      */
     private function parseBadRequest()
     {
-        if (isset($this->elements)) {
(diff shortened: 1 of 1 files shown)


calcinai:
Do you think it's safe to merge this into the current minor? I don't see an issue specifically, as things will be failing before that exception currently..

davidwindell:
Yep, I'd say so, it's only throwing an exception that could still be thrown in the same request so downstreams will already be handling this.

PR #341, opened 2018-02-26
setTotal, setSubTotal and setTotalTax added

medutis:
I've been using those properties to set total sums in my invoices directly.

### DIFF
--- a/src/XeroPHP/Remote/Response.php
+++ b/src/XeroPHP/Remote/Response.php
@@ -213,14 +213,14 @@ class Response
                 case 'ValidationErrors':
                     if (is_array($value)) {
                         foreach ($value as $error) {
-                            $this->element_errors[$element_index] = trim($error['Message'], '.');
+                            $this->element_errors[$element_index][] = trim($error['Message'], '.');
                         }
                     }
                     break;
                 case 'Warnings':
                     if (is_array($value)) {
                         foreach ($value as $warning) {
-                            $this->element_warnings[$element_index] = trim($warning['Message'], '.');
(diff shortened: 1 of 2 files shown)


src/XeroPHP/Remote/Response.php
```diff
@@ -213,14 +213,14 @@ public function findElementErrors($element, $element_index)
                 case 'ValidationErrors':
                     if (is_array($value)) {
                         foreach ($value as $error) {
-                            $this->element_errors[$element_index] = trim($error['Message'], '.');
+                            $this->element_errors[$element_index][] = trim($error['Message'], '.');
```
calcinai: I'm happy to merge the other functions, can you please clarify what's going on here?

medutis: Sorry to answer now, this shouldn't have been in the same PR, somehow I commited to the same.
But to answer that: sometimes there's more than one error for the same $element_index, and I need to see them all, not only the last out of two (let's say) errors.

calcinai: Will this cause issues of that property is being read as a string?

timacdonald: Even if this were to change, I think it would be best to consider this in a different PR. Keep history clean and PRs focused 👍

liamgcarter:
Hiya Guys.

I'm still learning git.
I need this code also. But do not want to over write the code in the vendor folder. As i wish to be able to update this extension when updates occur.

Can i use this pull request. Or do i need to extend this current class in my own application?

Regards

timacdonald:
@liamgcarter I'd say the simplest way is to just extend the class in your application and add this new code the the extended class.

Issue #352, opened 2018-03-14
Rate Limit Counter


olimortimer:
I'm struggling with the Xero Rate Limit, and unsure why as I'm not querying often enough to hit their limit.  Is there anything built into Xero-PHP to allow me to output the rate that I am querying at?  Ie a counter etc.

mogilvie:
Not that I am aware of.
You could create a log on each of your query execution, or more quickly, you view your endpoint history in the Xero developer portal.
Regards,
Mark

olimortimer:
Cheers @mogilvie

olimortimer:
Xero include a X-Rate-Limit-Problem in the HTTP headers, which would be good to retrieve back for more information on the rate limiting.

For example; `X-Rate-Limit-Problem: Daily` or `X-Rate-Limit-Problem: Minute`

I have a solution to output this, but it's messy...

PR #353, opened 2018-03-14
Include X-Rate-Limit-Problem in RateLimitExceededException message

olimortimer:
Includes X-Rate-Limit-Problem in RateLimitExceededException message as discussed here https://github.com/calcinai/xero-php/issues/352


calcinai:
I'm concerned that this change is not API compatible with previous versions and may need to be deferred until the next major, unless you're able to make your changes tot he response class with setters?

calcinai:
@olimortimer thoughts?

olimortimer:
Feel free to close if it isn't API compatible.

Issue #360, opened 2018-04-05
Property  TypeOfUnits of LeaveType should be a string


omgwhatuser:
Hi,

On line 123 of Models/PayrollAU/PayItem/LeaveType.php you have

`            'TypeOfUnits' => [true, self::PROPERTY_TYPE_FLOAT, null, true, false],
`

But as per Xero API docs it returns a string (ie. "Hours"), see JSON response excerpt below (extracted from https://developer.xero.com/documentation/payroll-api/payitems)

`
...
<LeaveType>
      <LeaveTypeID>f3a5f2f8-c14e-4223-ab58-4e779e0f6f2c</LeaveTypeID>
      <Name>Carer's Leave (unpaid)</Name>
     <TypeOfUnits>Hours</TypeOfUnits>
      <IsPaidLeave>false</IsPaidLeave>
      <ShowOnPayslip>false</ShowOnPayslip>
      <UpdatedDateUTC>2013-04-09T23:45:25</UpdatedDateUTC>
    </LeaveType>
...`

Is there any reasons a float[] was used for this property?

Regards,
Rafael

calcinai:
All of these models were generated using a doc scraper, feel free to update and make a PR!

PR #367, opened 2018-04-27
Issue #360 - TypeOfUnits should be a string and not float[]

omgwhatuser:
Please accept my changes.


calcinai:
Thanks for updating this

PR #372, opened 2018-05-03
Added missing parameters and functions

Healyhatman:
I made a different version of this pull request a while ago and it didn't happen. Still trying to learn how this github business works.

Your version of PayrollAU/Employee is still missing things (like Phone and more importantly TaxDelcaration) and the TaxDeclaration has incorrectly-typed or missing parameters. I cloned/downloaded your most recent version a few weeks back and broke everything on my site, which was a fun couple hours (100% my fault for not checking before pushing it to a live site) - I had used my own copies of a few of the classes but not all of them. Still no idea how composer works or how I could edit my code  here and have it put to the webserver other than zipping it up and uploading it manually. 

Anyways these are most of the fixed I can remember making.


calcinai:
Thanks for making this again.  The code looks fine, although can you please make sure it's `psr-2` and you're using spaces for indent not tabs (you can see this in the changes)

Re: downloading and swapping out the library, this is generally a Bad Idea (r) as there's no way to follow any semantic versioning.  I'd *strongly* recommend getting familiar with composer, that way you can be almost certain that you can update reliably without any API incompatible changes, and you can generally see the differences between versions

Healyhatman:
OK so I ran it through an online PSR-2 fixer, and then a convert-tabs-to-spaces thing, hopefully that's what you were after? 

I know you've answered this question somewhere else before me but is composer something I can use without CL access, on a shared server?

Healyhatman:
I don't have a "dev copy"? I just edit my scripts in notepad++ and upload
them with FileZilla. I honestly have no idea what you're talking about
haha. I'll look into it a bit once I finish my timesheets system. Once I
figure out why the dates are being stupid.

src/XeroPHP/Models/PayrollAU/Employee/TaxDeclaration.php
```diff
@@ -1,5 +1,5 @@
 <?php
-namespace XeroPHP\Models\PayrollAU\Employee;
+    namespace XeroPHP\Models\PayrollAU\Employee;
```
calcinai: I know it's fussy, but fix the indent in this one line and I'll merge! Thanks for the work

Healyhatman: Done

calcinai:
Oh I'm so sorry, I see we're somehow back to tabs for indents in a couple of those files. What IDE are you using? Does it have a function to fix formatting? In PHPStorm it's cmd+option+shift+L

Healyhatman:
I'm using notepad++ haha. I DID change it to use spaces instead of tabs

Healyhatman:
OK there we go. For some reason, when I edited the file with the tabbed namespace Github had it set to use size 8 tabs.

Healyhatman:
So @calcinai any other issues with it?

calcinai:
Looking good! Can you just clarify for me that property you deleted `HasTSLDebt`, did this used to exist? Why removed?

Healyhatman:
It used to exist, they replaced it with HasTradeSupportLoanDebt

calcinai:
@Healyhatman yep - the comments were auto generated.  If it was seen in a previous generation, then disappeared, the function was left there but marked deprecated for compatibility.

Are you about to chuck them back in and add that tag? I promise this is the last change!

Healyhatman:
The 'deprecated' comment is still in there, but you're saying put the methods and the HasTSLDebt attribute back in? Even though you can't use that attribute?

calcinai:
Only to maintain the semantic versioning - we need to keep it API compatible in case someone is calling it (albeit doing nothing)

Healyhatman:
OK There you are I put the property back in getProperties and I put the get/set methods back in. And I made sure it was using spaces and not tabs :D

Healyhatman:
Nothing else missing or incorrect @calcinai ?

PR #375, opened 2018-05-08
added new trait to allow Invoices to be emailed

wobinb:
The Xero API now allows for Invoices to be sent from the API itself. This PR adds this ability by adding a new trait to the Invoice model.

Usage:
$invoice->sendEmail();

### DIFF
--- a/src/XeroPHP/Traits/SendEmailTrait.php
+++ b/src/XeroPHP/Traits/SendEmailTrait.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace XeroPHP\Traits;
+
+use XeroPHP\Remote\Request;
+use XeroPHP\Remote\URL;
+use XeroPHP\Exception;
+
+trait SendEmailTrait
+{
+    public function sendEmail()
+    {
+        /**
+         * @var Object $this
+         */
+        $uri = sprintf('%s/%s/Email', $this::getResourceURI(), $this->getGUID());
+        
+        $url = new URL($this->_application, $uri);
+        $request = new Request($this->_application, $url, Request::METHOD_POST);
+        
+        $request->send();
+        
+        $response = $request->getResponse();
+        
+        if (false !== $element = current($response->getElements())) {
+            $attachment->fromStringArray($element);
+        }
+        
+        return $this;
+    }
+}
--- a/src/XeroPHP/Models/Accounting/Invoice.php
+++ b/src/XeroPHP/Models/Accounting/Invoice.php
@@ -4,6 +4,7 @@ namespace XeroPHP\Models\Accounting;
 use XeroPHP\Remote;
 use XeroPHP\Traits\PDFTrait;
 use XeroPHP\Traits\AttachmentTrait;
+use XeroPHP\Traits\SendEmailTrait;
 use XeroPHP\Models\Accounting\Invoice\LineItem;
 
 class Invoice extends Remote\Model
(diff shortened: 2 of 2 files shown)


calcinai:
Do you think this is something that's likely to be applied to other models? Also, do you think any validation is worthwhile here? (see below)

>The invoice must be of Type ACCREC and a valid Status for sending (SUMBITTED,AUTHORISED or PAID).

src/XeroPHP/Traits/SendEmailTrait.php
```diff
@@ -0,0 +1,31 @@
...
+         * @var Object $this
+         */
+        $uri = sprintf('%s/%s/Email', $this::getResourceURI(), $this->getGUID());
+        
+        $url = new URL($this->_application, $uri);
+        $request = new Request($this->_application, $url, Request::METHOD_POST);
+        
+        $request->send();
+        
+        $response = $request->getResponse();
```
Josh-G: Seems like this endpoint will never return content:
> The response code returned for a successful request is a 204 (no content).

Josh-G:
> Do you think this is something that's likely to be applied to other models?
Right now it seems that only invoices are part of the beta, but it is possible that Xero could add more email endpoints for entities such as quotes and receipts etc (https://help.xero.com/EmailTemplates)

> Also, do you think any validation is worthwhile here? (see below)
Seeing as most (all?) of the library relies on raising validation errors based on the response from Xero, I think it may be best to follow that standard in this case.

Josh-G:
It would also be nice to leave a docblock comment on it to? :)

wobinb:
Sorry for not getting back on here sooner.

> Do you think this is something that's likely to be applied to other models?

As noted, currently only the Invoices can be sent, but logically there are other documents that could be opened up for sending in the future.

> Also, do you think any validation is worthwhile here?

I'm not sure what you mean by validation? catching the problem before the API call is made? or parsing the response in the event of an error being returned from the API?

Forgive me, I'm not overly familiar with the library, how is this handled on other calls? 

Also, what is meant by a docblock?

calcinai:
@wobinb All valid points. Let's leave it up to the server for validation - have you tested the response you get with an invalid status?

I feel like [this](https://www.google.com/search?q=docblock) is an appropriate answer to explain a what a docblock is!

wobinb:
@calcinai I've added some comments to explain the function and a link to the documentation.

src/XeroPHP/Traits/SendEmailTrait.php
```diff
@@ -0,0 +1,31 @@
...
+        
+        if (false !== $element = current($response->getElements())) {
```
calcinai: Ah, is this just copied and pasted from the attachment trait? What does this serve in here?

calcinai: Reckon you could fix it up and I'll merge this in

DJLad97:
@phena109 Did you create any class that extended Invoices or did you put that function directly into the package's source code?

DJLad97:
@phena109 Did you create any class that extended Invoices or did you put that function directly into the package's source code?

phena109:
@DJLad97 Just a wrapper class in my project. Not exactly extending anything nor modifying the library. Hope this answer you question

src/XeroPHP/Traits/SendEmailTrait.php
calcinai: @wobinb can you please remove this line so I can merge this?

wobinb:
@calcinai Finally got my test environment backup to test, have removed the unnecessary lines.

PR #378, opened 2018-05-17
Standardise `setMethodName` return value

timacdonald:
Ahoy!

All of the models `setMethodName` methods return `$this` so you can chain them like a builder.

The only one that doesn't is `setApplication`. I know that this is not a bit different to the other setters - but I think adding it standardises the API as I personally was expecting to be able to chain...

`$invoice->asXeroInvoice()->setApplication($xeroClient)->getPDF();`

Like you can with other methods...

`$xeroInvoice->setContact($contact)->setDueDate($date)->....`

Where `$invoice` is the representation in my app. I can't see this being a breaking change for any apps myself.

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -78,10 +78,13 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
      * but will have to be like this for BC until the next major version.
      *
      * @param Application $application
+     * @return $this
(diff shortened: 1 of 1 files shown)


calcinai:
I don't see any harm in merging this now - beware - this will be deprecated in future versions as `$xero` will be required as a constructor parameter…

timacdonald:
No worries - future Tim can deal with that :)

PR #381, opened 2018-05-23
Add a method for for getting the response code.

dylanharbour:
I'd like to check that the response is a 200 in my implementation, but I cannot access the response property.

If I should be doing this another way, please advise. 

Thanks for the code! So far, it's working 💯 for me.


calcinai:
I see this being useful either way, although it's probably safest to just catch the Exception if it's not a `200`

PR #382, opened 2018-05-24
Update Item.php

theshelf:
Changed addSalesDetail and addPurchaseDetail to add value to a collection
Added setSalesDetails and setPurchaseDetails to set the entire collection

### DIFF
--- a/src/XeroPHP/Models/Accounting/Item.php
+++ b/src/XeroPHP/Models/Accounting/Item.php
@@ -1,38 +1,31 @@
 <?php
 namespace XeroPHP\Models\Accounting;
-
 use XeroPHP\Remote;
 use XeroPHP\Models\Accounting\Item\Purchase;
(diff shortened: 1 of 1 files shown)


calcinai:
Is this correct? It looks like you can end up with either a collection or a single object as this property...

calcinai:
Can I be picky and ask that you leave in the old functions but add the deprecated tag? They wouldn't have been working (I can't see how) anyway, so I think it's safe for them to be empty, but don't want to throw fatals.

theshelf:
You're correct - I had left the edited addSalesDetail/addPurchaseDetail functions when they aren't required at all - it's a switch from 

Have added a patch that removes the addSalesDetail/addPurchaseDetail functions and replaces them with the setSalesDetails/setPurchaseDetails.

calcinai:
Perfect, thanks.

PR #388, opened 2018-06-08
Add Exception to Invoice

mtimbs:
Error when trying to throw exception as class was not using it


calcinai:
Awesome, cheers for fixing that up.

PR #389, opened 2018-06-08
Instantiate LineItems if it's not set

mewejo:
Throws an E_NOTICE as the index LineItems isn't set by default


calcinai:
Thanks for this. I don't see any harm in adding this now, as it'd be throwing errors anyway if you weren't checking.

mewejo:
No worries - it was throwing an error for me. They could all do with having the same sort of check added to them though.

PR #391, opened 2018-06-22
Support HistoryRecords on Contacts

iangregory:
This PR adds support for HistoryRecords (a.k.a Notes) on Contacts in Xero. 

Xero API docs: https://developer.xero.com/documentation/api/history-and-notes

It should be fairly easy to extend this to the other objects which support HistoryRecords by adding the HistoryTrait to those models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,27 @@ XeroPHP
 [![Latest Stable Version](https://poser.pugx.org/calcinai/xero-php/v/stable)](https://packagist.org/packages/calcinai/xero-php)
 [![Total Downloads](https://poser.pugx.org/calcinai/xero-php/downloads)](https://packagist.org/packages/calcinai/xero-php)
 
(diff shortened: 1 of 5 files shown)


Josh-G:
The following models also support this trait
- BankTransactions
- BankTransfers
- Creditnotes
- ExpenseClaims
- Invoices
- Items
- Overpayments
- Payments
- Prepayments
- Purchase Orders
- Receipts
- Repeating Invoices

ShortsDev:
I'm very new to PHP and I'm slowly picking things up as I go along.

I've added the above code to include the history trait for contacts. But how do I actually make the call? So far I've made a call to get all the contact details and used the ->where to filter the results.

print "<pre>";
print_r($xero->load('Accounting\\Contact') -> execute());
print "<pre>";

My overall goal is to get this working for invoice history.

Any help would be greatly appreciated.

iangregory:
Some commits have incorrectly been associated with this PR. Will raise a new PR

PR #415, opened 2018-08-22
Expose x-rate-limit-problem header

### LINKED ISSUE #352: Rate Limit Counter

Josh-G:
Extension of #353 that maintains compatibility. This change adds it as a field rather than to the exception message.

Our use case is to monitor the number of merchants we are routinely hitting the rate limit for and expose this info to our support team.

Fixes #352.


Josh-G:
@calcinai any input?

calcinai:
@Josh-G sorry I'll take a loo later today. I see a couple of the function signatures have changed so need to make sure there are no implications of that.

calcinai:
_later today_

Josh-G:
@calcinai could I request a release for this please? 😄

PR #449, opened 2018-12-20
Add missing class property

timacdonald:
Hey! This class property declaration was missing. Added it.


calcinai:
Good spotting, thanks Tim

PR #451, opened 2018-12-20
Add missing doc types

timacdonald:
So were a bunch of doc blocks with missing parameter type information. I've added the type information.

I know that there are also a bunch of methods without doc blocks, but I wanted to fix up what was currently existing first.

I hope you don't mind me sending through some of these incremental changes. Trying to keep them small and focused for you so it isn't to much work to review them.

~There are some missing types in the `src/XeroPHP/Models/*` classes, but as they are generated I figured I see if I can fix that in the generator, rather than on the classes themselves.~

I just saw that you no longer use the generator. I'll leave this PR as it is and deal with the Models folder at a later date.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -100,7 +100,6 @@ abstract class Application
     /**
     * @param string $config
     * @param mixed $option
-    * @param mixed $value
     * @return mixed
     * @throws Exception
     */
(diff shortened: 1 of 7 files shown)


timacdonald:
gonna close these and do them class by class rather than spanning across the codebase like this.

PR #452, opened 2018-12-20
fix return types

timacdonald:
Continuing on with the doc block fixes - this time targeting incorrect return types.


timacdonald:
gonna close these and do them class by class rather than spanning across the codebase like this.

PR #453, opened 2018-12-21
Update code style rules

timacdonald:
Hey Michael,

Sorry for the burst of PRs, just had some time and wanted to send them through while I could.

Currently there are a bunch of different code styles throughout the project and I thought it might be good to lock down the preferred rules and enforce them to make the code more consistent, and hopefully by doing so, easier to maintain into the future.

## What I've done

1. Renamed the `.php_cs` to `.php_cs.dist`.

The `.php_cs` is meant to contain local rules specific to you, but the list version is supposed to be distributed with the package. [Config file docs](https://github.com/FriendsOfPHP/PHP-CS-Fixer#config-file). 

2. Explicitly specifies all rules to use, or not use.

I've gone through all the rules available and specified either yes, no, or specific configuration for each. [This awesome site](https://mlocati.github.io/php-cs-fixer-configurator/) allows you to visualise all the rules in action so you can see what they do.

Obviously there is going to be some of my own personal preferences in these rule specs, but I've generally opted for more strictness as I think the more the fixer can enforce, the better.

Some rules that I've not included have comments next to the explaining why they are not included, which is usually because they require a higher version of PHP or PHPUnit than this package supports. I think we should leave the comments there so if the package bumps PHP versions, you know what rules have become available. If there is no comment than it is probably just a personal preference.

3. Added the fixer cache file to the git ignore.

This should not be distributed, so I've ignored it.

## Will it break anything?

This PR does not actually change any library code. All I am doing is proposing the package adopts a specific set of rules around styling so they can be enforced in future PRs. Once we've decided on the rule set and this PR is merged (if you want it of course) then in follow up PRs I'd be looking to do the following:

1. Run each rule one-by-one and do a PR for each.

As you can imagine just running all the rules at once would create a nightmare of a PR that would be impossible to decipher. Instead I'd do a seperate PR for each individual rule. This means that you should be able to easily see what is changing and where it is being applied.

2. Add CS fixer to Travis

Once all rules have been run and merged, I'd then look at adding CS fixer to Travis CI to ensure all future PRs stick the the decided on rules.


calcinai:
LGTM!

PR #456, opened 2018-12-30
Add HistoryTrait to supported models

indygriffiths:
As per https://developer.xero.com/documentation/api/history-and-notes#SupportedDocs


calcinai:
@indygriffiths Cheers for that, I see there are a couple of whitespace changes in there, but I don't think it's worth another PR for them!

PR #459, opened 2019-01-05
Add php-cs-fixer as a dev dependency

timacdonald:
This PR adds the PHP Code Style fixer to the dev dependencies. As this is only a dev dependency it has no impact on the production composer requirements and is not a breaking change.

Having it as a dependency means you will be able to run it in C.I. once that is setup to utilise it and also run it locally with:

```
$ ./vendor/bin/php-cs-fixer fix
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "require-dev": {
         "phpunit/phpunit": "~4.7",
         "fabpot/goutte": "~2.0",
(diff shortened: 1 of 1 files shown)


timacdonald:
ahhh, failed as doesn't support 5.5; Will leave it out of this for now - you can just run it in CI instead of having it in the composer.json anyway, so that is not problem.

PR #460, opened 2019-01-05
Coding style: @PSR2

timacdonald:
This PR runs the PSR-2 style run over the codebase. I've run over this a couple of times but please scan yourself to make sure I haven't missed anything.

The only changes I can see are:

- Whitespace removal / addition
- Brace positioning
- Switch statement `// no break` comments
- Explicit `public` keyword on functions, instead of having it implicit.
- Breaking all method arguments onto new lines when already some on new lines.

These style fixes that I'll do consecutively will cause a slight headache for open PRs as we are probably going to touch a lot of the code, but hopefully it is the old bandaid trick: just rip it off and get it done with. Then we can put CI in place to ensure all future PRs respect the styling rules.


calcinai:
Thanks @timacdonald. This should mean that the PRs are a bit tidier

PR #465, opened 2019-01-16
Coding style: binary_operator_spaces

timacdonald:
Use one space on either side of binary operators.

See: https://github.com/FriendsOfPHP/PHP-CS-Fixer


timacdonald:
Hey @calcinai how do you feel about this rule?

Previously all the array `=>` thingo's (sorry for the technical jargon) were aligned vertically. Although this looks nice, it can mean that a PR to introduce another item in the array will need to touch every other item in the array, for example:

```php
return [
    'One'         => 1,
    'Two Hundred' => 200,
];
```

The to introduce another with a longer name and have them all line up you have to make changes to all the entries in the array.

```diff
return [
-    'One'         => 1,
-    'Two Hundred' => 200,
+    'One'                 => 1,
+    'Two Hundred'         => 200,
+    'Two Hundred and One' => 201,
];
```

So although you lose the visual appeal of everything being aligned, it makes future changes easier to consume.

What is more is that it also fixes a bunch of places where there is only one item in the array, but there is more than one space for no reason.

calcinai:
Makes sense to me!

Issue #468, opened 2019-01-17
saveAll does not reload InvoiceID (for invoices) - similar for others


cjsoftuk:
I've been investigating a saveAll issue.

If I call saveAll, the reload of data doesn't set replace to true.

https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/Model.php#L179

As a result, after a saveAll(), the GUIDs are not available.  This is a result of https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Application.php#L360

cjsoftuk:
Looking at this further, save() supports this behaviour, saveAll does not.  I'll attempt a patch.

PR #469, opened 2019-01-17
saveAll does not allow reload of IDs

cjsoftuk:
I think this addresses #468 - but needs to test it.  Any others experiencing the same issue, I'd welcome testing and feedback.


cjsoftuk:
I've tested this locally on my application that uses this library - seems to work OK for me - but would welcome maintainer feedback.

calcinai:
This has come up in the past and not sure why it  never got merged, thanks!

PR #472, opened 2019-01-22
Add error handling to documentation

timacdonald:
Adds some documentation around handling exceptions including a response code to exception map and also a brief example of how to handle an exception in PHP. Both provide links to more comprehensive and official resources.


calcinai:
I wonder if we should split up the readme with links to a docs folder?

timacdonald:
Probably a good idea once they are a bit more fleshed out that might be a good idea. Might feel a bit bare if we did that with the current docs (IMO).

You can also put content in expandable sections, so maybe that is a nicer approach for now.

I might spend some time on the docs in the next few days.

PR #482, opened 2019-02-07
Coding style: escape_implicit_backslashes

timacdonald:
Escape implicit backslashes in strings and heredocs to ease the understanding of which are special chars interpreted by PHP and which not.

See: https://github.com/FriendsOfPHP/PHP-CS-Fixer

**Note**: Some regex in this PR is changed, which naturally feels scary. But these are still equivalent. See / play here: https://3v4l.org/5COW1


timacdonald:
> I trust you

Famous last words 😉

calcinai:
That, and those regexes are only used for the (now defunct) doc parser.

PR #488, opened 2019-02-11
Fix repeating invoice -> schedule -> due date type

### LINKED ISSUE #257: Unable to get RepeatingInvoice->Schedule->DueDateType
I'm trying to inspect the schedule of a repeating invoice, and the DueDateType is getting dropped.

I'm first retrieving the repeating invoice:

```
$xero = new \XeroPHP\Application\PrivateApplication($config);
$schedule = $xero->loadByGUID('Accounting\\RepeatingInvoice', '...')->getSchedule();
```

If I dump this, I can see under `_application->oauth_client->request->response->elements` that the schedule details were successfully retrieved:

```
"Schedule" => array:6 [
    "Period" => "12"
    "Unit" => "MONTHLY"
    "DueDate" => "5"
    "DueDateType" => "OFFOLLOWINGMONTH"
    "StartDate" => "2016-04-01T00:00:00"
    "NextScheduledDate" => "2018-04-01T00:00:00"
]
```

However if I dump `$schedule->getDueDateType()` I am given an instance of `PaymentTerm` with empty `Bills` and `Sales` arrays. That's it. 

```
XeroPHP\Models\Accounting\Organisation\PaymentTerm {#779
  #_data: array:2 [
    "Bills" => null
    "Sales" => null
  ]
  ...
```

The DueDateType value of `OFFOLLOWINGMONTH` appears to be entirely lost. 

I also tried `$schedule->toStringArray()` and got this:

```
[
     "Period" => "12",
     "Unit" => "MONTHLY",
     "DueDate" => "5",
     "DueDateType" => [],
     "StartDate" => "2016-04-01",
     "NextScheduledDate" => "2018-04-01",
]
```

Shouldn't `DueDateType` just be a string response of `OFFOLLOWINGMONTH` in this case? Not sure why it's being treated as a `PaymentTerm`.

jszobody:
This PR fixes #257. 

The `DueDateType` specified in a repeating invoice schedule is provided as a string value from Xero.

In PHP the schedule looks like:

```php
"Schedule" => array:6 [
    "Period" => "12"
    "Unit" => "MONTHLY"
    "DueDate" => "5"
    "DueDateType" => "OFFOLLOWINGMONTH"
    "StartDate" => "2016-04-01T00:00:00"
    "NextScheduledDate" => "2019-03-01T00:00:00"
]
```

The problem is that the Schedule model uses `Organization\PaymentTerm` for the `DueDateType` property, resulting in a `PaymentTerm` instance with empty `Bills` and `Sales`. The string value "OFFOLLOWINGMONTH" is entirely lost.

It appears that `Organization\PaymentTerm` is the wrong model for a repeating invoice due date type. The nested `Bills` and `Sales` structure is used for Organizations and Contacts, but with a RepeatingInvoice Schedule it is a simple string.

I've already tested this PR and can successfully access the `DueDateType` property of my repeating invoices now.

Let me know if you'd prefer this to be an enum property with class constants. I wasn't sure how you'd want those constants named given the [lengthy payment term names](https://developer.xero.com/documentation/api/types#PaymentTerms). I also wasn't sure if you'd want the constants defined in the `Schedule` class that contains that `DueDateType` property, or if you'd rather put them in `Organization\PaymentTerm` even though we aren't using that model. And I figured since `Unit` was a string and not an enum I might get away with it. =)

### DIFF
--- a/src/XeroPHP/Models/Accounting/RepeatingInvoice/Schedule.php
+++ b/src/XeroPHP/Models/Accounting/RepeatingInvoice/Schedule.php
@@ -122,7 +122,7 @@ class Schedule extends Remote\Model
             'Period' => [false, self::PROPERTY_TYPE_INT, null, false, false],
             'Unit' => [false, self::PROPERTY_TYPE_STRING, null, false, false],
             'DueDate' => [false, self::PROPERTY_TYPE_INT, null, false, false],
-            'DueDateType' => [false, self::PROPERTY_TYPE_STRING, null, false, false],
+            'DueDateType' => [false, self::PROPERTY_TYPE_ENUM, null, false, false],
             'StartDate' => [false, self::PROPERTY_TYPE_DATE, '\\DateTimeInterface', false, false],
             'NextScheduledDate' => [false, self::PROPERTY_TYPE_DATE, '\\DateTimeInterface', false, false],
             'EndDate' => [false, self::PROPERTY_TYPE_DATE, '\\DateTimeInterface', false, false]
--- a/src/XeroPHP/Models/Accounting/RepeatingInvoice/Schedule.php
+++ b/src/XeroPHP/Models/Accounting/RepeatingInvoice/Schedule.php
@@ -3,7 +3,6 @@
 namespace XeroPHP\Models\Accounting\RepeatingInvoice;
 
 use XeroPHP\Remote;
-use XeroPHP\Models\Accounting\Organisation\PaymentTerm;
 
 class Schedule extends Remote\Model
 {
@@ -27,9 +26,9 @@ class Schedule extends Remote\Model
      */
 
     /**
-     * See Payment Terms
+     * Get the due date type
      *
-     * @property PaymentTerm DueDateType
+     * @property string DueDateType
      */
 
     /**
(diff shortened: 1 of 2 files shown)


jszobody:
The more I think about it the more I expect you'll want this to be an enum. So the question is where to put the constants.

```php
namespace  XeroPHP\Models\Accounting\Organisation;

class PaymentTerm {
    const OFFOLLOWINGMONTH = 'OFFOLLOWINGMONTH';
    ...
```

or

```php
namespace XeroPHP\Models\Accounting\RepeatingInvoice;

class Schedule {
    const DUE_DATE_TYPE_OFFOLLOWINGMONTH = 'OFFOLLOWINGMONTH';
```

or something different?

calcinai:
For consistency, I'd suggest the pseudo enum as you suggested (got this in an email but can't see here?)


```
class PaymentTerm {
    const OFFOLLOWINGMONTH = 'OFFOLLOWINGMONTH';
    ...
```

jszobody:
Roger that, will do. I had assumed as much, but then changed my mind after noticing that `Unit` with a value of "MONTHLY" is a string, not enum. 

I'll get this updated shortly.

jszobody:
I think that should do it. Confirmed working in my app. Let me know if I missed anything?

src/XeroPHP/Models/Accounting/RepeatingInvoice/Schedule.php
```diff
@@ -193,18 +192,18 @@ public function setDueDate($value)
...
         return $this->_data['DueDateType'];
     }
 
     /**
-     * @param PaymentTerm $value
+     * @param string $value
      * @return Schedule
      */
-    public function setDueDateType(PaymentTerm $value)
+    public function setDueDateType(string $value)
```
timacdonald:
```suggestion
    public function setDueDateType($value)
```

I would remove this type-hint as the package supports 5.5+ and scalar type-hints are only available 7.0+.

I know that you are removing a type hint, which means nobody on a lower version of PHP is probably using this, but they also might be hitting an error and not reporting it. So maybe we remove this while making this change?

Then we can add type hints everywhere later down the line once we bump versions?

jszobody: Good call, done!

PR #501, opened 2019-02-17
Code style: non_printable_character

timacdonald:
Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other invisible unicode symbols.


calcinai:
Ha, all from the docs. I'm guessing they've got something funny in whatever they use to admin those pages.

PR #508, opened 2019-02-17
Code style: phpdoc_order

timacdonald:
Annotations in PHPDoc should be ordered so that @param annotations come first, then @throws annotations, then @return annotations.


calcinai:
@timacdonald Sorry I broke this.

timacdonald:
No problem! I'll take a look during the week. It's gonna happen when I'm throwing so many at ya.

timacdonald:
@calcinai just rebased - should be sweet to merge now.

PR #523, opened 2019-02-28
Check my facade to make easy to use it

ezequiel9:


calcinai:
I'm sorry but this PR doesn't really align with the project.  Have you looked at [/amochohan/xerolaravel](https://github.com/amochohan/xerolaravel)?

ezequiel9:
It would have been great to know before !. I lost a lot of time with that
.. thanks for the information .. 😑

timacdonald:
@ezequiel9 it isn't that there is something wrong with your PR as an idea, it is that **this** package has absolutely no relation to Laravel - but your PR assumes the code will be used in Laravel, as it is using Laravel specific classes and functionality.

You should submit your PR to a **Laravel specific** wrapper of this package, as @calcinai suggested.

PR #547, opened 2019-03-27
Code style: strict_comparison

timacdonald:
Comparisons should be strict.


calcinai:
Ew. How did I let those through?

PR #548, opened 2019-03-27
Code style: strict_param

timacdonald:
Functions should be used with `$strict` param set to `true`.

I just made this (https://3v4l.org/YFIk8) to make sure any values, like booleans, are actually converted correctly with this change, as it is risky, but it looks like it is sweet to me.


calcinai:
Cool

PR #553, opened 2019-03-28
add tracking option id

nimmen:
add tracking option id

### DIFF
--- a/src/XeroPHP/Models/Accounting/TrackingCategory.php
+++ b/src/XeroPHP/Models/Accounting/TrackingCategory.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace XeroPHP\Models\Accounting;
 
 use XeroPHP\Remote;
@@ -6,7 +7,6 @@ use XeroPHP\Models\Accounting\TrackingCategory\TrackingOption;
 
 class TrackingCategory extends Remote\Model
 {
-
     /**
      * The Xero identifier for a tracking categorye.g. 297c2dc5-cc47-4afd-8ec8-74990b8761e9
      *
@@ -237,7 +237,7 @@ class TrackingCategory extends Remote\Model
     public function addOption(TrackingOption $value)
     {
         $this->propertyUpdated('Options', $value);
-        if (!isset($this->_data['Options'])) {
+        if (! isset($this->_data['Options'])) {
             $this->_data['Options'] = new Remote\Collection();
         }
         $this->_data['Options'][] = $value;
--- a/src/XeroPHP/Models/Accounting/TrackingCategory.php
+++ b/src/XeroPHP/Models/Accounting/TrackingCategory.php
@@ -1,5 +1,4 @@
 <?php
-
 namespace XeroPHP\Models\Accounting;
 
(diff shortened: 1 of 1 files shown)


src/XeroPHP/Models/Accounting/TrackingCategory.php
```diff
@@ -1,12 +1,12 @@
 <?php
-
```
timacdonald: Add new line back in

src/XeroPHP/Models/Accounting/TrackingCategory.php
```diff
@@ -1,12 +1,12 @@
...
 namespace XeroPHP\Models\Accounting;
 
 use XeroPHP\Remote;
 use XeroPHP\Models\Accounting\TrackingCategory\TrackingOption;
 
 class TrackingCategory extends Remote\Model
 {
+
```
timacdonald: Remove new line

src/XeroPHP/Models/Accounting/TrackingCategory.php
```diff
@@ -236,7 +237,7 @@ public function getOptions()
     public function addOption(TrackingOption $value)
     {
         $this->propertyUpdated('Options', $value);
-        if (! isset($this->_data['Options'])) {
+        if (!isset($this->_data['Options'])) {
```
timacdonald: restore space after NOT (!) operator

src/XeroPHP/Models/Accounting/TrackingCategory.php
```diff
@@ -262,4 +263,24 @@ public function setOption($value)
         $this->_data['Option'] = $value;
         return $this;
     }
+
+    /**
+     * @return string $value
+     * Return tracking option id
```
timacdonald: Think this docblock should just be adjusted to the return type.

```
/**
 * @return string
 */
```

timacdonald:
I'm not seeing this value available on the root TrackingCategory in the docs: https://developer.xero.com/documentation/api/tracking-categories

Could you provide a screenshot of a response that contains this value on the root object and not in the Options array?

src/XeroPHP/Models/Accounting/TrackingCategory.php
nimmen: Can adjust if needed, but there are other examples, so tried to make this more consistent.

Existing case:
    /**
     * @return string $value
     * Returns selected option name
     */
    public function getOption()

nimmen:
Added back whitespace formatting.
In our case, we always get Options as NULL, not sure if this may be api version related, but it seems redundant/not working, or might be related to client setup?
Sending screenshot with actual issue(before my changes), just pushing this upstream so we would not have problems later, when updating. After my changes everything works well for us now.

Untitled

nimmen:
As you can see we get back tracking option name, but not tracking option id.
What else seems redundant/not working:
"TrackingCategoryName" which is "Name"
"TrackingOptionName" which is "Option"
"Options" which is always NULL in my test cases
"Status" ? don't think it's working

Issue #559, opened 2019-04-10
twig < 1.38 has a security vulnerability 


buggedcom:
https://symfony.com/blog/twig-sandbox-information-disclosure

Can you please bump your require version.

timacdonald:
Just opened a PR to address this: https://github.com/calcinai/xero-php/pull/560

PR #560, opened 2019-04-10
Remove deprecated generator

timacdonald:
As the generator is no longer used, and you will always have it in version control, I'm proposing we just ditch it from the repo.

I know when I first accessed this library I actually thought the models were still generated automatically from the generator. So removing it may help clear that up for potential new contributors.

It also removes the `--dev` dependencies [`twig/twig`](https://github.com/twigphp/Twig) and [`fabpot/goutte`](https://github.com/FriendsOfPHP/Goutte). Now although these are only meant for development, and should not be install in production, if you are using something like [`roave/security-advisories`](https://github.com/Roave/SecurityAdvisories) it will error out and not let you install this package. The reason being, [as mentioned here](https://github.com/calcinai/xero-php/issues/559), the Twig version required has a security vulnerability.

As this is really internal code, I don't see this as a breaking change.


calcinai:
Thanks Tim. I was about to do exactly the same thing.

timacdonald:
No worries at all

calcinai:
Sad (but happy) to see it go!

timacdonald:
forever in our ~hearts~ git history

PR #574, opened 2019-05-15
PaymentService.php

Abhayraj7ingh:
This file is used to support Payment Service APIs.


Josh-G:
[API documentation](https://developer.xero.com/documentation/api/payment-services)
We've used a very similar custom model in the past.

Abhayraj7ingh:
> [API documentation](https://developer.xero.com/documentation/api/payment-services)
> We've used a very similar custom model in the past.

PR #535, opened 2019-03-25
Feature/payroll uk

dextermb:
As requested [here](https://github.com/calcinai/xero-php/issues/322#issuecomment-475912713).

_Ignore the package JSON renaming. Instead of installing it under my own package I now just overwrite the repository_

PR #578, opened 2019-05-22
update model.php to define default format for the body of POST or PUT…

wobinb:
… calls

Newer API endpoints only accept the body to be in JSON format. By adding this function, the existing models will default to XML, and allow new models to be defined with json instead. 

This would allow Application.php to be modified in the future to submit json when necessary.

I appreciate this might not be the tidiest solution and am open to more elegant solutions.

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -558,4 +558,10 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
     {
         unset($this->_data[$offset]);
     }
+
+    public static function getBodyFormat()
+    {
+        return 'xml';
+    }
+
 }


wobinb:
Never mind, just seen that this has been tackled for the the UK Payroll PR....

calcinai:
@wobinb thanks for the work on this. I haven't switched it to default behaviour as I haven't had a good chance to test the json parsing yet.

When I originally implemented it, it was the other way around–json was only supported on a few endpoints.

wobinb:
@calcinai As it stands new endpoints are JSON only, older endpoints can use either, and AU Payroll is XML only.

I quite like the way the it has been done here https://github.com/calcinai/xero-php/pull/535 by omitting the RootNodeName value.

It might also be worth considering how to handle the Projects endpoints which have POST and PUT the other way round - So POST is for creating an item, and PUT for an edit...

Issue #335, opened 2018-01-31
Unable to add more than one allocation to overpayment


hilobok:
When I'm trying to add the `Allocation` to `Overpayment` which already has allocations, I got a validation exception on saving overpayment. As far as I understand [Xero API](https://developer.xero.com/documentation/api/overpayments#PUT) waits for only new allocations instead of all previous besides new ones.

Sample code:

```php
<?php

// overpayment already has allocations
$overpayment->addAllocation((new Overpayment\Allocation())
  ->setInvoice($invoice)
  ->setAppliedAmount($amount)
);

$application->save($overpayment);
```

got this exception

```
    "exception": {
        "class": "XeroPHP\\Remote\\Exception\\BadRequestException",
        "message": "A validation exception occurred (Overpayment cannot be allocated as amount is greater than amount outstanding)\nValidation errors:\nThe invoice you are trying to allocate to has already been paid off",
        "code": 10,
    }
```

As workaround I'm creating allocations directly via API, like this:

```php
<?php

    private function allocate(Overpayment $overpayment, Invoice $invoice, $amount)
    {
        $allocation = (new Overpayment\Allocation())
            ->setInvoice($invoice)
            ->setAppliedAmount($amount)
        ;

        $url = new URL($this->xero, sprintf(
            '%s/%s/%s',
            $overpayment::getResourceURI(),
            $overpayment->getOverpaymentID(),
            $allocation::getResourceURI())
        );

        $request = new Request($this->xero, $url, Request::METHOD_PUT);
        $request->setBody(Helpers::arrayToXML([
            'Allocations' => [
                $allocation->toStringArray(),
            ],
        ]));
        $request->send();
    }
```

gurubobnz:
Can confirm this - to test allocating an overpayment to invoices I was allocating a dollar at a time to an invoice. The first allocation sent one, the second sent two (overpayment now had three allocations on it), the third sent three (overpayment now had six on it) and so on.

This terribly naive PR will fix it, but probably breaks the collection somehow https://github.com/calcinai/xero-php/compare/master...gurubobnz:gurubobnz-dont-append-to-allocations

Josh-G:
Does it work if you remove all items from the allocations collection first? Based on the original code sample:

```php
$overpayment->getAllocations()->removeAll();

$overpayment->addAllocation((new Overpayment\Allocation())
  ->setInvoice($invoice)
  ->setAppliedAmount($amount)
);

$application->save($overpayment);
```

[...]

PR #582, opened 2019-06-05
When adding an allocation replace allocations with a single allocation.

gurubobnz:
The Xero API is additive. You can only add new allocations, not append to an existing collection. Fix for https://github.com/calcinai/xero-php/issues/335 but needs more work before merging.

### DIFF
--- a/src/XeroPHP/Models/Accounting/Overpayment.php
+++ b/src/XeroPHP/Models/Accounting/Overpayment.php
@@ -582,7 +582,7 @@ class Overpayment extends Remote\Model
         if (! isset($this->_data['Allocations'])) {
             $this->_data['Allocations'] = new Remote\Collection();
         }
-        $this->_data['Allocations'][] = $value;
+        $this->_data['Allocations'] = [$value];
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


calcinai:
@gurubobnz What more work do you think is required before merging?

gurubobnz:
@calcinai I've confirmed that this code here works also https://github.com/calcinai/xero-php/issues/335#issuecomment-498990082 so need to rework this PR to do something similar. I'll give it a crack.

gurubobnz:
Have tidied up the PR.

timacdonald:
How would you add multiple *new* allocations in the one request with this?

```php
$overpayment->addAllocation($allocation_1);

$overpayment->addAllocation($allocation_2);
```

This would result in *only* `$allocation_2` being added right? Because `$allocation_1` would be removed when you add the second one?

I've never dealt with overpayment allocations so I'm not sure if anyone would ever even do this, but something to consider.

gurubobnz:
Hmm yes you're right, that would not work. I will have a think about this.

calcinai:
@gurubobnz any thoughts?

PR #587, opened 2019-07-01
Support for a callback on successful API call

gurubobnz:
We use this to track usage of the Xero API in our application so we can manage to stay under the 5,000 calls per day. Usage (for us):

```
    $xero->onRequestCompleted(function($response, $info, $headers){
        $log = new Apilog;
        $log->System = 'Xero';
        $log->URL = $info['url'];
        $log->HttpCode = $info['http_code'];
        $log->ResponseSize = strlen($response);
        $log->save();
    });
```

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -451,9 +451,9 @@ abstract class Application
         return $object;
     }
 
-    public function requestCompleted($response) {
+    public function requestCompleted($request) {
         foreach($this->requestCompletedCallbacks as $callback) {
-            call_user_func($callback, $response);
+            call_user_func($callback, $request);
         }
     }
 
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -56,6 +56,11 @@ abstract class Application
     protected static $_type_config_defaults = [];
 
     /**
+     * @var Closure
+     */
+    protected $requestCompletedCallback;
+
+    /**
      * @param array $config
      */
     public function __construct(array $config)
@@ -446,16 +451,14 @@ abstract class Application
         return $object;
     }
 
-	protected $requestCompletedCallback;
+    public function requestCompleted($response, $info, $headers) {
+        if($this->requestCompletedCallback instanceof \Closure) {
+            call_user_func($this->requestCompletedCallback, $response, $info, $headers);
+        }
+    }
 
-	public function requestCompleted($response, $info, $headers) {
-		if($this->requestCompletedCallback instanceof \Closure) {
-			call_user_func($this->requestCompletedCallback, $response, $info, $headers);
-		}
-	}
-
-	public function onRequestCompleted($callback) {
-		$this->requestCompletedCallback = $callback;
-	}
+    public function onRequestCompleted($callback) {
+        $this->requestCompletedCallback = $callback;
+    }
 
 }
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -56,9 +56,9 @@ abstract class Application
     protected static $_type_config_defaults = [];
 
     /**
-     * @var Closure
+     * @var array
      */
-    protected $requestCompletedCallback;
+    protected $requestCompletedCallbacks = [];
 
     /**
      * @param array $config
@@ -452,13 +452,15 @@ abstract class Application
     }
 
     public function requestCompleted($response, $info, $headers) {
-        if($this->requestCompletedCallback instanceof \Closure) {
-            call_user_func($this->requestCompletedCallback, $response, $info, $headers);
+        foreach($this->requestCompletedCallbacks as $callback) {
+            if($callback instanceof \Closure) {
+                call_user_func($callback, $response, $info, $headers);
+            }
         }
     }
 
     public function onRequestCompleted($callback) {
-        $this->requestCompletedCallback = $callback;
+        $this->requestCompletedCallbacks[] = $callback;
     }
 
 }
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -453,9 +453,7 @@ abstract class Application
 
     public function requestCompleted($response, $info, $headers) {
         foreach($this->requestCompletedCallbacks as $callback) {
-            if($callback instanceof \Closure) {
-                call_user_func($callback, $response, $info, $headers);
-            }
+            call_user_func($callback, $response, $info, $headers);
         }
     }
 
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -445,4 +445,17 @@ abstract class Application
 
         return $object;
     }
+
+	protected $requestCompletedCallback;
+
+	public function requestCompleted($response, $info, $headers) {
+		if($this->requestCompletedCallback instanceof \Closure) {
+			call_user_func($this->requestCompletedCallback, $response, $info, $headers);
+		}
+	}
+
+	public function onRequestCompleted($callback) {
+		$this->requestCompletedCallback = $callback;
+	}
+
 }
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -451,9 +451,9 @@ abstract class Application
         return $object;
     }
 
-    public function requestCompleted($response, $info, $headers) {
+    public function requestCompleted($response) {
         foreach($this->requestCompletedCallbacks as $callback) {
-            call_user_func($callback, $response, $info, $headers);
+            call_user_func($callback, $response);
         }
     }
 
--- a/src/XeroPHP/Remote/Request.php
+++ b/src/XeroPHP/Remote/Request.php
@@ -125,9 +125,9 @@ class Request
         }
 
         $this->response = new Response($this, $response, $info, $headers);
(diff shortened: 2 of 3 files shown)


src/XeroPHP/Application.php
```diff
@@ -445,4 +450,15 @@ public function delete(Remote\Model $object)
...
         return $object;
     }
+
+    public function requestCompleted($response, $info, $headers) {
+        if($this->requestCompletedCallback instanceof \Closure) {
+            call_user_func($this->requestCompletedCallback, $response, $info, $headers);
+        }
+    }
+
+    public function onRequestCompleted($callback) {
```
timacdonald: Maybe this should be called `onRequestSuccess` as `completed` could be interpreted as success AND failed requests.

timacdonald:
I dig this. Nice way to open up the request to devs - as it is currently hard for them to dig in and see the raw response

gurubobnz:
@timacdonald Thanks for the comments, nice suggestion. I've added to this PR. I was wondering also if the response, info, and headers should be passed by reference, which would allow the callback to modify them if required.

```
call_user_func($callback, &$response, &$info, &$headers);
```

I suspect the best answer is no, keep it simple. Thoughts?

timacdonald:
My opinion: lets not pass by reference.

Although I can understand there could be come value in this for certain scenarios, I can also see it causing headaches not only for the developers, but also for @calcinai with support requests saying "The status code is 200" when someone decided to change the value from 500 to 200 or something silly like that.

But this is really @calcinai call :)

src/XeroPHP/Application.php
```diff
@@ -445,4 +450,17 @@ public function delete(Remote\Model $object)
 
         return $object;
     }
+
+    public function requestCompleted($response, $info, $headers) {
+        foreach($this->requestCompletedCallbacks as $callback) {
+            if($callback instanceof \Closure) {
```
timacdonald: I think we can get rid of this `instanceof` check. It opens it up to all types of callables (i.e. array based) and I think it is better to let PHP error out and say "You passed me something this can't handle" rather than just silently skipping it.

Thoughts?

calcinai:
@timacdonald  
>but also for @calcinai with support requests 

You deal with some of this more than I do these days! But I agree, I don't see any reason to silently  modify the response.

src/XeroPHP/Application.php
gurubobnz: Gotcha, [this commit](https://github.com/calcinai/xero-php/pull/587/commits/598393ec0d13cd96ecad85cc3a28816b2f3e451e) fixes that. Thanks :)

timacdonald: no worries. Cool PR - should come in handy for a lot of people I think.

gurubobnz:
I realised today there's a slightly better way to do this by passing the response object itself instead of the parts that make it up, that way you can get to the original request as well. I will propose this tomorrow.

gurubobnz:
Sorry still iterating on a final solution here ... if this PR is too messy let me know and I'll close it and make a new one ^_^

I decided to have the original `$request` passed to the `onRequestCompleted` method as this already has a getter for the response which means I don't have to add a `getRequest()` method to the response itself. The request already has a `getResponse()` method.

Only slight concern is that by doing this we are opening the door to the other methods on the request/response being usable in ways not intended, for example the send() method on the request could be called again in an `onRequestCompleted()` callback and it may appear that you could use this to handle API cooldowns which is not the intention (at least not at this stage).

With the latest commit the usage example would then be:
```
    $xero->onRequestCompleted(function($request){
        
        echo implode("\n", [
            'Request:',
            '   Method: '.$request->getMethod(),
            '   Headers: '.count($request->getHeaders()),
            '   URL: '.$request->getURL()->getFullURL(),
            'Response:',
            '   Status: '.$request->getResponse()->getStatus(),
            '   Headers: '.count($request->getResponse()->getHeaders()),
            '   ResponseBody: '.substr($request->getResponse()->getResponseBody(), 0, 40).' ...',
            ''
        ]);
    }
```
and the output of the above would be:
```
Request:
   Method: GET
   Headers: 2
   URL: https://api.xero.com/api.xro/2.0/Invoices/9d6170af-1a81-4cfb-8081-c99efce20f59
Response:
   Status: 200
   Headers: 13
   ResponseBody: <Response xmlns:xsd="http://www.w3.org/2 ...
```
Much tidier I think.

gurubobnz:
Sorry still iterating on a final solution here ... if this PR is too messy let me know and I'll close it and make a new one ^_^

I decided to have the original `$request` passed to the `onRequestCompleted` method as this already has a getter for the response which means I don't have to add a `getRequest()` method to the response itself. The request already has a `getResponse()` method.

Only slight concern is that by doing this we are opening the door to the other methods on the request/response being usable in ways not intended, for example the send() method on the request could be called again in an `onRequestCompleted()` callback and it may appear that you could use this to handle API cooldowns which is not the intention (at least not at this stage).

With the latest commit the usage example would then be:
```
    $xero->onRequestCompleted(function($request){
        
        echo implode("\n", [
            'Request:',
            '   Method: '.$request->getMethod(),
            '   Headers: '.count($request->getHeaders()),
            '   URL: '.$request->getURL()->getFullURL(),
            'Response:',
            '   Status: '.$request->getResponse()->getStatus(),
            '   Headers: '.count($request->getResponse()->getHeaders()),
            '   ResponseBody: '.substr($request->getResponse()->getResponseBody(), 0, 40).' ...',
            ''
        ]);
    }
```
and the output of the above would be:
```
Request:
   Method: GET
   Headers: 2
   URL: https://api.xero.com/api.xro/2.0/Invoices/9d6170af-1a81-4cfb-8081-c99efce20f59
Response:
   Status: 200
   Headers: 13
   ResponseBody: <Response xmlns:xsd="http://www.w3.org/2 ...
```
Much tidier I think.

gurubobnz:
Sorry still iterating on a final solution here ... if this PR is too messy let me know and I'll close it and make a new one ^_^

I decided to have the original `$request` passed to the `onRequestCompleted` method as this already has a getter for the response which means I don't have to add a `getRequest()` method to the response itself. The request already has a `getResponse()` method.

Only slight concern is that by doing this we are opening the door to the other methods on the request/response being usable in ways not intended, for example the send() method on the request could be called again in an `onRequestCompleted()` callback and it may appear that you could use this to handle API cooldowns which is not the intention (at least not at this stage).

With the latest commit the usage example would then be:
```
    $xero->onRequestCompleted(function($request){
        
        echo implode("\n", [
            'Request:',
            '   Method: '.$request->getMethod(),
            '   Headers: '.count($request->getHeaders()),
            '   URL: '.$request->getURL()->getFullURL(),
            'Response:',
            '   Status: '.$request->getResponse()->getStatus(),
            '   Headers: '.count($request->getResponse()->getHeaders()),
            '   ResponseBody: '.substr($request->getResponse()->getResponseBody(), 0, 40).' ...',
            ''
        ]);
    }
```
and the output of the above would be:
```
Request:
   Method: GET
   Headers: 2
   URL: https://api.xero.com/api.xro/2.0/Invoices/9d6170af-1a81-4cfb-8081-c99efce20f59
Response:
   Status: 200
   Headers: 13
   ResponseBody: <Response xmlns:xsd="http://www.w3.org/2 ...
```
Much tidier I think.

timacdonald:
Although I agree getting the response as a `Response` object would be a nicer approach - I believe the original implementation was the better approach (totally just my opinion).

The `parse` method just before the callbacks are run is our problem.

It throws a bunch of exceptions if the request is not 200-a-ok:

- BadRequestException
- UnauthorizedException
- NotFoundException
- InternalErrorException
- NotImplementedException
- OrganisationOfflineException
- RateLimitExceededException
- NotAvailableException

So therefore it is possible that the request "completed" but was not "successful" and thus the callbacks will not be executed.

It also means that the Xero library has touched the response content before the developer gets a chance to inspect the payload.

I wonder if we introduce two callback hooks:

- `onRequestCompleted`
- `onRequestSuccess`

I'm not convinced they are the best pairing of names, so feel free to improve them.

Completed gets the raw response content before the library has touched it. Success gets the Request object after it has been parsed.

Thoughts?

gurubobnz:
@timacdonald It's all good, thanks for the comments :)

Yeah I didn't realise there were a possible bunch of exceptions thrown - and I found myself wanting to pass the curl response itself in to the callback so I could get the original request, when I realised that hey that's what we have already right here as an object ...

I don't want to blow this feature up into more than it needs to be, but I do like the idea of `onRequestCompleted` and `onRequestSuccess`, the naming being similar to that of jQuery's ajax methods (`complete` and `success`). This does also imply though that there should be support for an `onRequestFailed` also.

I don't mind writing the code for this if it improves the quality of the API. I'll hold off until we get some input from @calcinai because it has to be right to be merged.

Thanks again for your input.

gurubobnz:
@calcinai Thoughts on this please? I'm keen to wrap this PR up :)

calcinai:
@gurubobnz I think those three hooks make sense, if you're happy to update the PR that should be enough.

gurubobnz:
Given the work that has been done on the OAuth2 branch it makes sense to look at this in that context, good discussion above, thanks all, I will propose a new PR when I get my head around it.

PR #594, opened 2019-07-04
Code style: phpdoc_indent

timacdonald:
Docblocks should have the same indentation as the documented subject.


timacdonald:
@calcinai fixed the conflicts here.

PR #595, opened 2019-07-04
Code style: phpdoc_separation

timacdonald:
Annotations in PHPDoc should be grouped together so that annotations of the same type immediately follow each other, and annotations of a different type are separated by a single blank line.


timacdonald:
@calcinai fixed the conflicts here.

PR #610, opened 2019-07-31
Add CurrentRecord getters and setters to every PayItem model.

wbercx:
I ran into some trouble with timesheet integrations where users had marked used EarningsRates as Inactive in Xero.

I found out that PayItems [now have a `CurrentRecord` attribute](https://developer.xero.com/documentation/payroll-api/payitems#EarningsRates) that indicates whether an item is active. 

This PR adds a getter and setter for `CurrentRecord` to every PayItem model so you can handle deactivated items.

### DIFF
--- a/src/XeroPHP/Models/PayrollAU/PayItem/DeductionType.php
+++ b/src/XeroPHP/Models/PayrollAU/PayItem/DeductionType.php
@@ -106,6 +106,7 @@ class DeductionType extends Remote\Model
             'ReducesTax' => [true, self::PROPERTY_TYPE_FLOAT, null, false, false],
             'ReducesSuper' => [true, self::PROPERTY_TYPE_STRING, null, false, false],
             'DeductionTypeID' => [false, self::PROPERTY_TYPE_STRING, null, false, false],
+            'CurrentRecord' => [false, self::PROPERTY_TYPE_STRING, null, false, false]
         ];
     }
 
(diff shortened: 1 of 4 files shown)


timacdonald:
I know it is inconsistent with other similar values in these classes, but I wonder if these should be `PROPERTY_TYPE_BOOL` and return the boolean value rather than just `"true"` or `"false"` as strings.

calcinai:
@wbercx What are your thoughts on the above?

wbercx:
Booleans sound great and intuitive. It's how I initially thought the library would work.

I then found the getters I care about return `"true"` and `"false"`, and just remembered "xero-php returns stringified booleans". That seems consistent in the parts of the library I use myself (admittedly limited in scope). I like that consistency because it is predictable and easy to work with.

I'll treat the above as 2 votes for going with booleans. I should be able to make the necessary change later this week.

In my opinion, we should at some point make this consistent across the entire library.

Healyhatman:
I JUST came here to have a sook about this, I need this please :)

calcinai:
@wbercx How'd you get on with these changes? I have a major change coming up so we could probably release as part of that.

wbercx:
I've made some changes, but will have to test them tomorrow morning to see if anything else is required. Flying blind at the moment.

Apologies for the delay.

calcinai:
@wbercx You're the last person to be apologising!

PR #621, opened 2019-08-15
Generate nonce with random bytes rather than time

timacdonald:
This PR changes the way a nonce is generated. Previously it was dependent on time and it seems there have been multiple attempts to stop hitting collisions in previous PRs

e.g. https://github.com/calcinai/xero-php/pull/174

But it still seems to be an issue.

I propose we move to use open SSL random bytes to generate the nonce. As they do not need to be cryptographically secure, and open SSL is already a requirement for the library, I figure it is probably a good move.

It would be great if I could get some testing done with this as I actually haven't tested this yet - hence the "draft" status.

### DIFF
--- a/src/XeroPHP/Remote/OAuth/Client.php
+++ b/src/XeroPHP/Remote/OAuth/Client.php
@@ -222,24 +222,33 @@ class Client
     }
 
     /**
-     * Generic nonce generating function for the request.
-     * It's important that it's long enough as the spec says the
-     * server will reject any request that reuses one.
-     *
-     * @param int $length
+     * This snippet was taken from implementations of the laravel/framework
+     * \Illuminate\Suppport\Str::random() method.
      *
      * @return string
      */
-    private function getNonce($length = 20)
+    private function getNonce()
     {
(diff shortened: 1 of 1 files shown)


calcinai:
Man, I feel like this has been changed many times.  How is it STILL using time to generate a nonce? That's silly.

timacdonald:
I think this is ready for the big time. Some notes about the PR:

- I've removed the description in the docblock. I personally felt it was not required as it is essentially describing what a nonce is. I don't feel it needs there, but happy to put it back if you think it is handy.
- The method is used in once place in the library, and the method is private. As it is only called in once place and the length of the nonce is never specified elsewhere, I've removed the `$length` parameter.
- The method now uses `random_bytes` in PHP 7+ and falls back to `openssl_random_pseudo_bytes` on PHP 5. This does not need to be cryptographically secure so openssl shouldn't be a concern, but I think it is a good idea to rely on something that is built in, rather than an extension, when possible. Open SSL is currently a requirement of this package.
- I've left a comment in there about how the method does not need to be cryptographically secure, just for future readers that are aware that `openssl_random_pseudo_bytes` is not cryptographically secure.

PR #623, opened 2019-08-18
add nonce generation tests

timacdonald:
Some tests to compliment: https://github.com/calcinai/xero-php/pull/621


timacdonald:
The failed build doesn't seem to be related to this PR. I think there is an issue perhaps with Travis and PHP 5.5 or something.

Issue #637, opened 2019-09-19
Updating/Archiving Tracking Category Options


tysonarmstrong:
Hi,

I'm trying to archive a large number of tracking category options. It appears it isn't possible to save() the TrackingOption object directly, so I've also tried updating the parent Tracking Category options collection with the updated option (ie. setting Status to ARCHIVED). When I try to save this Tracking Category I get a 404. 
The endpoint it's posting to is:
https://api.xero.com/api.xro/2.0/TrackingCategories/{TrackingCategoryID}/Options/
which is giving the 404.

The API docs say that you can update a Tracking Option directly at:
https://api.xero.com/api.xro/2.0/TrackingCategories/{TrackingCategoryID}/Options/{TrackingOptionID}
but the TrackingOption class has no supported methods.
(https://developer.xero.com/documentation/api/tracking-categories)

Am I messing this up somewhere, or is it not supported by this API client?

Cheers,
Tyson

kgdiem:
Duplicate of #256 

Same here

PR #640, opened 2019-09-25
Update TrackingOptions to conform w/ API docs

### LINKED ISSUE #256: Adding TrackingOption to TrackingCategory issue
Hi, 
I have issue to add TrackingOption to TrackingCategory by using TrackingCategory::addOption() function.
After I spent some time on it I found in the XeroPHP\Models\Accounting\TrackingCategory\TrackingOption file, the values getResourceURI() and getRootNodeName() return don't match the Xero API request url and body:
`PUT https://api.xero.com/api.xro/2.0/TrackingCategories/{TrackingCategoryID}/Options`
```
<Options>
  <Option>
    <Name>New Option</Name>
  </Option>
  <Option>
    <Name>Another New Option</Name>
  </Option>
</Options>
```

```
/**
     * Get the resource uri of the class (Contacts) etc
     *
     * @return string
     */
    public static function getResourceURI()
    {
        return 'TrackingOptions';
    }


    /**
     * Get the root node name.  Just the unqualified classname
     *
     * @return string
     */
    public static function getRootNodeName()
    {
        return 'TrackingOption';
    }
```
I changed them to
```
/**
     * Get the resource uri of the class (Contacts) etc
     *
     * @return string
     */
    public static function getResourceURI()
    {
        return 'Options';
    }


    /**
     * Get the root node name.  Just the unqualified classname
     *
     * @return string
     */
    public static function getRootNodeName()
    {
        return 'Option';
    }
```
, and now it works. Not sure it may cause some other issues or not, but it temporarily fixes the issue.

kgdiem:
Was broken, resolves #256, #637

### DIFF
--- a/src/XeroPHP/Models/Accounting/TrackingCategory/TrackingOption.php
+++ b/src/XeroPHP/Models/Accounting/TrackingCategory/TrackingOption.php
@@ -31,7 +31,7 @@ class TrackingOption extends Remote\Model
      */
     public static function getResourceURI()
     {
-        return 'TrackingOptions';
+        return 'Options';
     }
 
     /**
@@ -41,7 +41,7 @@ class TrackingOption extends Remote\Model
      */
     public static function getRootNodeName()
     {
-        return 'TrackingOption';
+        return 'Option';
     }
 
     /**


calcinai:
@kgdiem I think this is a risky one, as the endpoint names are different in different contexts.  Which entity have you aligned these to?

kgdiem:
To TrackingCategories:
https://developer.xero.com/documentation/api/tracking-categories#PostOptions

Maybe there can be different models for other contexts, like `BankAccount\TrackingCategory\TrackingOption`, but where applicable?

StoyanPenev:
Hi @calcinai, 
could you please prioritise on this pull, I do not see anything risky about it, its currently broken and not working. The context for TrackingOption (this is what is being changed) is only in the context of a XeroPHP\Models\Accounting\TrackingCategory so there are no different contexts of usage. 
Could you get this merged, currently I have to fix this manually in the vendor, its ok as far as I am on local env but I am going to need this once I get to production env.

Thank you :)

calcinai:
Ok I've merged this in, can you please test out on master? I'll do some more testing myself

PR #645, opened 2019-10-09
Expose DateUTC with getDateUTC getter function

Drewyn:


calcinai:
I wonder why this was not there all along?

PR #655, opened 2019-11-19
Fixing Item Code being required for products

### LINKED ISSUE #404: Must manually set Code property to dirty to save Items
To update an item Xero requires the Code property to be passed even if you have the GUID of the item.  The only way I can see to update an existing item where I am not making a change to the Code property is to manually set the Code property to dirty.

jay-oswald:
This fixes #404 which I had an issue with myself when using the API.

The Item code is required for updating an item in the Xero API.

This is stripped out automatically by the dirty checker if you're updating an item because is hasn't changed, which throws an error.

The workaround I have put in place for my current project is just doing the below, to force the dirty flag to be set.
```
$item->setCode('abc');
$item->setCode($code);
```

In this Pull Request I have added a new function to the Remote Model, where it checks a property to see if it is required. If a property is class as required then it does not check if it's dirty or not, it always includes it.

I wrote it in a way so that the `isRequired` function can be easily extended if there are any other endpoints that require it

It looks like it may be required for Invoices, as LineItems, Contact and Type are all required, but I have not tested this yet as I am not currently using Invoices

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -247,7 +247,7 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
             }
 
             //if we only want the dirty props, stop here
-            if ( ! $this->isNeeded($property) && $dirty_only && ! isset($this->_dirty[$property])) {
+            if ( ! $this->isRequired($property) && $dirty_only && ! isset($this->_dirty[$property])) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


calcinai:
@jay-oswald thanks for your contribution.  I think this can be achieved without defining the isRequired function, but by referencing the class' properties array. The first parameter for 'Code' in this case is true, denoting mandatory.

Check out the model class and you shoulder what I mean.

jay-oswald:
Just had a look at the code, thats a much better solution then what I did, I'll update the PR when I get a chance

calcinai:
@jay-oswald Any update on this?

jay-oswald:
This project got put on hold for a while.

Just updated it, kept the is required function, to keep the `toStringArray` function if statement cleaner.

Make sure the property exists, and that the first element in the array is true.

PR #663, opened 2020-01-02
Fix for current() returning null when used with Remote\Collection

OJMichael:
I noticed credit note/overpayment allocations were not being sent to Xero, with the following message appearing in the logs:

> Warning Error: get_class() expects parameter 1 to be object, null given in [/var/www/app/Vendor/calcinai/xero-php/src/XeroPHP/Application.php, line 384]
> 
> Error: Class name must be a valid object or a string
> #0 /var/www/app/Vendor/calcinai/xero-php/src/XeroPHP/Application.php(273): XeroPHP\Application->savePropertiesDirectly()

After some investigation, I noticed the `current()` function when used with a `Remote\Collection` class always returns null. I replaced this with a `offsetGet(0)` method call on the object, and this appears to be working correctly.

For reference, I'm using PHP 7.4.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -408,7 +408,8 @@ abstract class Application
                 //Then actually save
                 $property_objects = $object->$property_name;
                 /** @var Remote\Model $property_type */
-                $property_type = get_class(current($property_objects));
+                $first_property = $property_objects->offsetGet(0);
+                $property_type = get_class($first_property);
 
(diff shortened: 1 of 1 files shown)


calcinai:
@OJMichael How does this behave if you've manipulated the collection (or just created it with a non-zero index)

OJMichael:
> @OJMichael How does this behave if you've manipulated the collection (or just created it with a non-zero index)

Fair point, it would have required an item in the "0th" index.

 I've reverted to the original `current()` function but using the `getArrayCopy()` method on the `Remote\Collection`.

This is working correctly for me, and will always use the first item in the Collection regardless of index.

PR #671, opened 2020-01-28
Fix an error in Response when Xero returns a 204

tobz-nz:
They do not return a `Content-Type` header with their `240` responses.

e.g. the send email endpoint returns a `204`


calcinai:
@tobz-nz I've just pushed an update that slightly tidies up that whole block, if you could update your copy and test, that'd be great.

tobz-nz:
Yeah it was a pretty quick dirty fix haha. Will check it out

PR #676, opened 2020-02-04
Update EarningsRate.php to include AllowanceType property

Healyhatman:
Creating an Allowance in Xero requires or at least prefers an undocumented "AllowanceType" property.
This update adds the getters and setters, and constants for the possible values of the property.


calcinai:
Looks good, but can you please format to psr-2

Healyhatman:
Done, I think?

PR #682, opened 2020-02-07
Guzzle exception fix

Yendall:
Exceptions weren't been captured and parsed correctly when a Bad Response Exception was thrown by the server, resulting in a truncated message from Guzzle rather than a parsed exception

### DIFF
--- a/src/XeroPHP/Remote/Request.php
+++ b/src/XeroPHP/Remote/Request.php
@@ -88,30 +88,6 @@ class Request
     }
 
     /**
-     * Parse Response from Guzzle / HTTP Client
-     * @param $response
-     * @throws Exception
-     * @throws Exception\BadRequestException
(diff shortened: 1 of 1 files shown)


calcinai:
I think this should still throw an exception, right?

calcinai:
Oh I see, it'll throw the exception when parsed.  That could be simplified to just catch the exception then set `$guzzleResponse = $e->getResponse()` then let it fall out of the exception.

Yendall:
> Oh I see, it'll throw the exception when parsed. That could be simplified to just catch the exception then set `$guzzleResponse = $e->getResponse()` then let it fall out of the exception.

Oh yep that makes sense, I'll test that out, push and see what you think

calcinai:
Looks good to me.  Thanks for the work.

PR #699, opened 2020-03-17
Fix 2 dates being returned as strings prop

Healyhatman:
expectedPaymentDate and plannedPaymentDate are supposed to be DateTimes not Strings


calcinai:
Thanks Kamron!

PR #700, opened 2020-03-17
Removed redundant forward slash separating base_url and path

reggietheroman:
Getting
`https://api.xero.com//api.xro/2.0/BrandingThemes`
instead of
`https://api.xero.com/api.xro/2.0/BrandingThemes`

### DIFF
--- a/src/XeroPHP/Remote/URL.php
+++ b/src/XeroPHP/Remote/URL.php
@@ -96,6 +96,6 @@ class URL
      */
     public function getFullURL()
     {
-        return sprintf('%s/%s', $this->base_url, $this->path);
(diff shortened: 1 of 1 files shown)


reggietheroman:
We could always just remove the trailing slash from the config base_url though. Really don't know what would be better in this situation tbh.

calcinai:
@reggietheroman My preference would probably be to remove it  from he base config. Are you able to see here else it is in use?

reggietheroman:
>  Are you able to see where else it is in use and 

ill look into it.

> confirm this won't introduce issues?

this is harder but it will have to be done. might have to add some test/s just to make sure.

calcinai:
@reggietheroman How'd you go?

PR #701, opened 2020-03-19
Add setPayslipMessage function

Healyhatman:
Missing method to set a PayRun's payslip message
I have tested this by adding the method, setting the message, saving, and verifying the change on Xero


calcinai:
Thanks @Healyhatman

PR #707, opened 2020-03-26
Handle null guids in loadByGuid and loadByGuids

paprikati:
If we don't do this, then if the $guid property is null, we collect a page of models and then end up returning a random model. This behaviour is very unexpected to a consumer of this library

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -188,6 +188,10 @@ class Application
         /** @var Remote\Model $class */
         $class = $this->validateModelClass($model);
 
+        if(!$guid){
+            throw new Remote\Exception\NotFoundException;
+        }
+
         $uri = sprintf('%s/%s', $class::getResourceURI(), $guid);
         $api = $class::getAPIStem();
 
@@ -224,6 +228,10 @@ class Application
     {
         /** @var $class Remote\Model */
         $class = $this->validateModelClass($model);
+
+        if(empty($guids)){
+            return [];
+        }
 
         $uri = sprintf('%s', $class::getResourceURI());
         $api = $class::getAPIStem();


calcinai:
@paprikati thanks for the PR.  Since this is effectively covering programmer error, it is probably useful to tint the `array` type and use `empty()` in `loadByGUIDs()` so it actually handles all cases.

With the singular loading, which cases are we trying to handle?  Just `null`, or null and empty strings? or both and invalid GUIDs? Same as above, it could be a little bit more explicit as it's arguably not the responsibility of this library to check that the API is being used as expected.

Not trying to be negative, these are just the things that come to mind.  Thanks again

calcinai:
@paprikati Does the above sound sensible?

paprikati:
Sorry, will take a look properly tomorrow morning (UK time)

paprikati:
> @paprikati thanks for the PR. Since this is effectively covering programmer error, it is probably useful to hint the array type and use empty() in loadByGUIDs() so it actually handles all cases.

Agree.

> With the singular loading, which cases are we trying to handle? Just null, or null and empty strings? or both and invalid GUIDs? Same as above, it could be a little bit more explicit as it's arguably not the responsibility of this library to check that the API is being used as expected.

We are trying to make sure we never return a random resource, which is the unexpected and dangerous behaviour. invalid GUIDs are fine - we pass them through to Xero, Xero returns an error, all good. So I think we need to catch empty string and null, which the code does as-is?

Will update to use Empty() now

calcinai:
@paprikati Thanks for that, makes sense

maheshemvigo:
We are trying to integrate Xero into our CRM. Currently we are trying to do the same in our local. Our CRM is in PHP Laravel.

We are using the SDK https://github.com/calcinai/xero-php> . 

We were able to get the connection initially using the customer ID and the refresh token. Then Using the access token, we were able to access the data. 

But after 30 min when the validity of the access token is expired, we used refresh token to renew the access token, but at that point of time the data becomes null. Could you please help us with what needs to be done exactly?

PR #709, opened 2020-03-28
Whitelisted the GUID property when using toStringArray($dirty_only = true)

edegaudenzi:
Discussion:
while the decision to force $dirty_only = true in the Application::saveAll() method should be reported into the documentation somewhere, it might be beneficial to take the following consideration:
The code already knows if the SaveAll is going to be a PUT or a POST: $has_guid = true; decides it.

So, since the GUID number functions de facto as a primary key, why don't we insert the GUID in the XML by default, if present? In this way the $dirty_only functionality remains untouched and developers headache is gone! ;)


calcinai:
Thanks @edegaudenzi

PR #712, opened 2020-04-01
Fix Guzzle exception wrapping

### LINKED ISSUE #711: Call to a member function getBody() on null
I'm almost at the end of migrating to OAuth 2.0, all I need to do now is interact with the API to get things. However running ->load(model)->execute() gives me the exception in the title. It also happens if I manually build and send a Remote\Request.

This is what my manual request looks like before I try to send it
```
XeroPHP\Remote\Request {#2081 ▼
  -app: XeroPHP\Application {#2071 ▼
    #config: array:1 [▼
      "xero" => array:4 [▼
        "base_url" => "https://api.xero.com/"
        "core_version" => "2.0"
        "payroll_version" => "1.0"
        "file_version" => "1.0"
      ]
    ]
    -transport: GuzzleHttp\Client {#2069 ▼
      -config: array:8 [▼
        "headers" => array:3 [▶]
        "handler" => GuzzleHttp\HandlerStack {#2068 ▶}
        "allow_redirects" => array:5 [▼
          "max" => 5
          "protocols" => array:2 [▼
            0 => "http"
            1 => "https"
          ]
          "strict" => false
          "referer" => false
          "track_redirects" => false
        ]
        "http_errors" => true
        "decode_content" => true
        "verify" => true
        "cookies" => false
        "idn_conversion" => true
      ]
    }
  }
  -url: XeroPHP\Remote\URL {#2065 ▼
    -base_url: "https://api.xero.com/"
    -endpoint: "PayItems"
    -path: "payroll.xro/1.0/PayItems"
  }
```

Josh-G:
resolves #711 (probably - based on the info available)

In https://github.com/calcinai/xero-php/pull/682 a fix was made to ensure that bad responses were caught by catching Guzzle exceptions. However, the catch is too broad and catches non-response related failures (`ConnectTimeout` etc), which do not have a wrapped `Response` object in the exception.

We should only catch `BadResponseException`s and allow Guzzle to throw other exceptions as necessary. Alternatively, we can start wrapping these other Guzzle exceptions into `XeroPHP\Remote\Exception`. I've chosen the former as 1.x didn't throw exceptions in those conditions.

I've added a test for the above behaviour. 
I've also bumped the phpunit version constraint (still 5.6 compatible) as there were no blockers to doing so. Let me know if you want me to split that into a separate PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
       "ext-simplexml": "*"
    },
    "require-dev": {
-      "phpunit/phpunit": "~5"
+      "phpunit/phpunit": "^5"
    },
    "autoload": {
       "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
       "ext-simplexml": "*"
    },
    "require-dev": {
-      "phpunit/phpunit": "~4.7"
+      "phpunit/phpunit": "~5"
    },
    "autoload": {
       "psr-4": {
--- a/src/XeroPHP/Remote/Request.php
+++ b/src/XeroPHP/Remote/Request.php
@@ -108,7 +108,7 @@ class Request
 
         try {
             $guzzleResponse = $this->app->getTransport()->send($request);
-        }  catch (\GuzzleHttp\Exception\GuzzleException $e) {
+        }  catch (\GuzzleHttp\Exception\BadResponseException $e) {
             $guzzleResponse = $e->getResponse();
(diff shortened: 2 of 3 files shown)


Josh-G:
@calcinai fyi - it seems that the travis build is no longer setting the github check status

calcinai:
@Josh-G will take a look at that.  Thanks for this.

PR #732, opened 2020-04-27
Fix throwing RateLimitExceededException

### LINKED ISSUE #731: Rate Limit Exceeded
Rate Limit Exceed has an error code of 429 but this isn't being picked up by the Response class.

I think some code needs updating to catch this error such as the following

const STATUS_RATE_LIMIT_EXCEEDED = 503;
to
const STATUS_RATE_LIMIT_EXCEEDED = 429;

/calcinai/xero-php/src/XeroPHP/Remote/Response.php

Josh-G:
resolves #731 

This adds support for the new 429 response code in the oauth2 API whilst maintaining backwards compatibility with the oauth1 API. Additionally, the refactor to Guzzle caused the casing of headers to change - breaking the parsing of the rate limit info headers.

Exposes the new Retry-After header.

Tests included


calcinai:
Nice work! Thanks.

PR #738, opened 2020-05-09
Move LineItem into the Accounting Namespace

mikerockett:
This PR does the following:

1. Creates a new LineItem class in the Accounting namespace. All documented properties included (hope I didn't miss anything 🙈)
2. Deprecates all previous LineItem classes by removing their functionality and extending the new LineItem class. Deprecation docblocks have been added to each of these.
3. All classes that use line items will now use the new LineItem class when converting collection instances to string.
4. The LINEAMOUNT_TYPE_EXCLUSIVE, LINEAMOUNT_TYPE_INCLUSIVE, and LINEAMOUNT_TYPE_NOTAX constants in the Invoice model are deprecated, and the new TYPE_EXCLUSIVE, TYPE_INCLUSIVE, and TYPE_NOTAX constants on the LineItem model should be used instead.

Opening as a draft PR as I'd like to do some local testing to make sure we're all good here.


mikerockett:
I need to make some updates here to remove the version constraint from the deprecation notices - as a new version has since been released.

mikerockett:
Version constraints have been removed from the line item deprecation notices. :-)

Issue #739, opened 2020-05-12
The save() method doesn't appear to throw an exception when status code 429 (too many requests) is returned


scottjs:
Hi,

I'm trying to batch insert some invoices and I'm finding that the save() method doesn't throw an exception when 429 response is returned:

```
try {
    $response = $xeroInvoice->save();
} catch (NotFoundException $exception) {
    Log::info('we dont get here');
} catch (RateLimitExceededException $exception) {
    Log::info('or here');
} catch (\Exception $e) {
    Log::info('or even here');
}

// If I've hit a request limit, this will return 429.
Log::info($response->getStatus());
```

For me to handle this properly I'm having to check for a 200:

```
if ($response->getStatus() !== 200) {
  // terminate
}

// continue with next steps
```

Am I missing something, is there a better way? 

Thanks,
Scott.

calcinai:
@scottjs are you using OAuth2? I _think_ it still throws 503s for OAuth1, butt hat should still be being caught in your code.  Do you see 429 in `$response->getStatus()`?

scottjs:
Hey :) Thanks for the reply!

I am using OAuth2 and I do eventually see a 429 from `$response->getStatus()` after I've hit the API too many times. 

I'm wondering if there should a check for this status code in the `parse()` method in `Response.php`?

scottjs:
Hey :) 

I've just noticed this PR: https://github.com/calcinai/xero-php/pull/732

This is exactly what I need to solve my problem as it's causing me quite a few issues, mainly because I'm creating a contact in the system if it doesn't already exist. 

Problem is that the query to check if a contact already exists may return `null` (for not found) or `null` if the API returns 429.

Without it throwing an exception my code will then think that the contact doesn't exist and tries to create it, so there's a chance it will create duplicate contacts. 

When do you think you'll be able to release the code in the approve PR?

scottjs:
Fixed by https://github.com/calcinai/xero-php/pull/732 and https://github.com/calcinai/xero-php/pull/740

PR #740, opened 2020-05-13
Throw exception on non-2xx response

scottjs:
Something I would have found useful while working with this library is if the response class threw an exception when a non-2xx status code is returned from the API. 

In the case of my issue https://github.com/calcinai/xero-php/issues/739 I realise this has already been fixed as part of https://github.com/calcinai/xero-php/pull/732, however had the response class still returned a fallback exception on a non-200, this could saved a couple of headaches. 

Currently the [Xero Documentation](https://developer.xero.com/documentation/api/http-response-codes) shows a specific set of status codes, however it's clear that the Xero API can change what is returned at any time without documenting it.

So, in the event that an unsuccessful undocumented status code is returned that is not being handled by this library, the response class will assume the request was successful. 

When this happens, any consuming code can't easily detect if there was a request failure or if the query just returned no results. 

An improvement to this change could be to check for an array of success codes (201, 202, etc) just in case Xero ever decide to return these in the future, although these aren't currently documented. 

Would be interested to hear your thoughts. 

Test included.

### DIFF
--- a/src/XeroPHP/Remote/Exception/UnknownStatusException.php
+++ b/src/XeroPHP/Remote/Exception/UnknownStatusException.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace XeroPHP\Remote\Exception;
+
+use XeroPHP\Remote\Exception;
+
+class UnknownStatusException extends Exception
+{
+    //
+}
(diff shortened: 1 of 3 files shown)


Josh-G:
I know that the API can also return `204 No Content` for some types of requests.

An exception should only be thrown if it's not 2xx code, in my opinion.

scottjs:
> I know that the API can also return `204 No Content` for some types of requests.
> 
> An exception should only be thrown if it's not a 2xx code, in my opinion.

Yep agreed, thanks for the reply. I have updated the PR.

calcinai:
> however it's clear that the Xero API can change what is returned at any time without documenting it.

@scottjs  Welcome to the wonderful world of the Xero API! This looks good though, thanks.

scottjs:
> > however it's clear that the Xero API can change what is returned at any time without documenting it.
> 
> @scottjs Welcome to the wonderful world of the Xero API! This looks good though, thanks.

Haha, indeed :) No worries, thanks for merging.

PR #750, opened 2020-07-13
Update Query.php

Healyhatman:
Add the createdByMyApp flag
Used in Accounting => Invoices and probably other places I haven't checked


calcinai:
Thanks @Healyhatman

PR #757, opened 2020-07-18
Report detail on unknown responses and intercept a potential scope issue

tonybyng:


calcinai:
Thanks @tonybyng - can you please update your PR to be PSR-2?

tonybyng:
Will do. Sorry. By the way, assigning arrays to class constants came in 5.6 rather than 5.5. Because my editor reads the PHP requirements from the composer.json file it keeps highlighting the array constant at the top of the Response.php as an error but if I change to 5.6 then all is good. You might want to bump the min requirements to PHP 5.6 unless there is a specific reason to stay at 5.5 and in which case that would need refactoring. Haven't checked if there is a similar situation elsewhere

calcinai:
@tonybyng good spotting - I've replaced that with a static [in this commit](cc223c382869efd0397ff2065afea2eafad24947)

The 5.5 version requirement is so it works on Google AppEngine

PR #694, opened 2020-03-06
Add Particulars, Code, Reference property

michgeek:
Add Particulars, Code, Reference as per batch payments documentation 

https://developer.xero.com/documentation/api/batch-payments#PUT

PR #764, opened 2020-09-08
Add Particulars, Code, Reference property

michgeek:
Same PR as #694

Add Particulars, Code, Reference as per batch payments documentation
https://developer.xero.com/documentation/api/batch-payments#PUT


calcinai:
Thanks @michgeek

PR #768, opened 2020-09-29
Initialise model $_data array to null

jeremyj11:
When creating and retrieving models eg PayrollAU\Employee various properties may not be set, resulting in undefined index errors when calling getters.
Eg PayrollAU\Employee::getBankAccounts() will throw an undefined index error if $_data['BankAccounts'] is not set:

```
/**
 * @return BankAccount[]|Remote\Collection
 */
 public function getBankAccounts()
 {
     return $this->_data['BankAccounts'];
 }
```
Rather than modifying multiple models to check that the property is set before returning, this pull request ensures all available properties are initialised to null upon model creation.

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -76,6 +76,7 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
         $this->_dirty = [];
         $this->_data = [];
         $this->_associated_objects = [];
+        $this->fromStringArray([]);
     }
 
     /**


Healyhatman:
Will this cause a problem though when saving a model? Are there endpoints that will see this null property and assume you mean to delete it?

calcinai:
@Healyhatman that was my initial thought. Will need to test.

Healyhatman:
It would definitely mess up PayrollAU / PayItems. If you tried to update just one type, it would include the null-filled properties for the rest and then delete them all.

From https://developer.xero.com/documentation/payroll-api/payitems

> Note: Deleting payroll PayItems works as follows. If you include a PayItem type (example: <DeductionTypes>) you must include each <DeductionType>, otherwise it will be deleted. If you omit the PayItem type from your request all PayItems of that type will not be modified or deleted.

jeremyj11:
ahh sorry my bad! Didn't realise things deleted with null.

The only other option seems to be using a lot of issets, eg changing
```
/**
     * @return BankAccount[]|Remote\Collection
     */
    public function getBankAccounts()
    {
        return $this->_data['BankAccounts'];
    }
```
to
```
/**
     * @return BankAccount[]|Remote\Collection
     */
    public function getBankAccounts()
    {
        return isset($this->_data['BankAccounts']) ? $this->_data['BankAccounts'] : null;
    }
```
For my use case I need to do this in 2 getter functions in PayrollAU / Employee, but I see this getter pattern used everywhere in nearly all models.. I'm wondering how no one else has encountered this problem before..

What I'm doing:
1) Getting an employee, creating if not exists
2) Checking if Employee pay templates and bank accounts exist, updating if yes, creating if no

Step 2: Calling  Employee::getPayTemplate() or Employee::getBankAccounts() will fail will if the Employee model was newly created as opposed to being  retrieved from Xero.

I could of course detect a newly created Employee and not call the getter functions but ideally the getter functions should not throw an error regardless of model state.. anyway I'll test more and come up with a non breaking solution, my apologies for the premature PR !

calcinai:
@jeremyj11 you could always do `isset($employee->BankAccounts)`?

jeremyj11:
ahh sorry my bad! Didn't realise things deleted with null.

The only other option seems to be using a lot of issets, eg changing
```
/**
     * @return BankAccount[]|Remote\Collection
     */
    public function getBankAccounts()
    {
        return $this->_data['BankAccounts'];
    }
```
to
```
/**
     * @return BankAccount[]|Remote\Collection
     */
    public function getBankAccounts()
    {
        return isset($this->_data['BankAccounts']) ? $this->_data['BankAccounts'] : null;
    }
```
For my use case I need to do this in 2 getter functions in PayrollAU / Employee, but I see this getter pattern used everywhere in nearly all models.. I'm wondering how no one else has encountered this problem before..

What I'm doing:
1) Getting an employee, creating if not exists
2) Checking if Employee pay templates and bank accounts exist, updating if yes, creating if no

Step 2: Calling  Employee::getPayTemplate() or Employee::getBankAccounts() will fail will if the Employee model was newly created as opposed to being  retrieved from Xero.

I could of course detect a newly created Employee and not call the getter functions but ideally the getter functions should not throw an error regardless of model state.. anyway I'll test more and come up with a non breaking solution, my apologies for the premature PR !

PR #769, opened 2020-09-29
Update PayTemplate.php

Healyhatman:
SuperLines and LeaveLines were not retrieved as arrays as they should've been


calcinai:
Thanks @Healyhatman

Issue #786, opened 2020-12-15
fromStringArray method is not working as expected


tiaguinhor:
Hey guys, I'm using the up to date calcinai `2.2.4` but the fromStringArray method from `Model` class is not working as expected.

When passing the second argument as `true`, it should override all the values as the previous versions (Calcinai 1), but that method is only changing `_data` and not `_dirty` as well, so it's not impacted on the API. 
![image](https://user-images.githubusercontent.com/9930297/102279263-33f4a680-3f7f-11eb-87a5-b0c0cdc74e23.png)

If you check any setter, the methods are calling `propertyUpdated` for updating the `_dirty` property and then the `_data`:
![image](https://user-images.githubusercontent.com/9930297/102279083-f2fc9200-3f7e-11eb-812c-4a448125b48c.png)

So updating the `Model` for doing the same, solved the problem and worked like the previous versions:
![image](https://user-images.githubusercontent.com/9930297/102279145-0b6cac80-3f7f-11eb-97ba-44754840bd23.png)

Right now we are overriding that method in our own class, but hope it helps others and also a fix could be released soon for that.

Thank you!

PR #787, opened 2020-12-15
Add propertyUpdated on fromStringArray method

tiaguinhor:
https://github.com/calcinai/xero-php/issues/786

### DIFF
--- a/src/XeroPHP/Models/Accounting/Organisation.php
+++ b/src/XeroPHP/Models/Accounting/Organisation.php
@@ -950,4 +950,12 @@ class Organisation extends Remote\Model
     {
         return $this->_data['OrganisationID'];
     }
+
+    /**
+     * @return string
+     */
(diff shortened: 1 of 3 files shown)


calcinai:
@tiaguinhor thanks for the PR.  Just thinking about this, I'm worried it may change some behaviour if this method is being used to load cached data within an application (or something similar).

Do you think adding another parameter to the function would be an acceptable solution to maintain backward compatibility?

tiaguinhor:
Hey @calcinai, I think the second argument was already designed on that purpose. But yes, a third argument would be a solution too. I see on the repository we have some TDD, none of them tests those behaviours? Btw, I didn't run them. Cheers, mate.

calcinai:
@tiaguinhor adding that propertyUpdated method there will force all requests to contain the entire object, and I'm not sure that is desirable.

PR #794, opened 2021-02-04
Implement Query->addWhere() Date support 

jennph:
Update Remote Query->addWhere()
Add PhpUnit Test for Query->where()
Update Readme to give date filtering example

--------------------------

'Date' is an optimised filtering parameter for several Xero Accounting models (e.g. [Invoices](https://developer.xero.com/documentation/api/invoices#optimised-parameters)), but

`$xero->load(Invoice::class)->where('Date', 'DateTime(2020,11,25)') `

fails with a Fatal Error:

`Fatal Error: Uncaught XeroPHP\Remote\Exception\BadRequestException: Operator '==' incompatible with operand types 'DateTime?' and 'String' ()`

This pull request updates the Remote Query addWhere function to recognise 'DateTime(...)' as a second parameter, and format it without enclosing quotes.

The addWhere function already has special cases to handle ints, floats, boolean values, and guids, so this is a two line change following the same format.

It also adds a unit test and updates the readme to give valid Date filtering examples.

**Additional Information**

Query has three date related functions: date(DateTime dt), fromDate(DateTime dt), toDate(DateTime dt)

These add parameters directly to the rest query url, rather than into the where clause of the url.
These parameters are not used in the 2.0 api for the Accounting Models that I have used. Invoice and Contact queries use date parameters within their where filters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Search for objects meeting certain criteria
 $xero->load(Invoice::class)
     ->where('Status', Invoice::INVOICE_STATUS_AUTHORISED)
     ->where('Type', Invoice::INVOICE_TYPE_ACCREC)
-    ->where('Date'), 'DateTime(2020,11,25)'
+    ->where('Date', 'DateTime(2020,11,25)')
     ->execute();
 
 $xero->load(Invoice::class)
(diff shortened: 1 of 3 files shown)


calcinai:
Looks great, thanks.  One improvement I could see would be to make the where clause just accept a DateTime object, and remove the need for the regular expression etc.

Issue #797, opened 2021-03-15
empty(XeroPHP\Remote\Collection) always returns false


edegaudenzi:
Empty \ArrayObject(s) are not evaluated correctly by PHP empty() method:
E.g.
`$arr = new \ArrayObject();`
`var_export(empty($arr));`

&#35; false, but the human would expect true.

PR #798, opened 2021-03-15
#797 implemented the XeroPHP\Remote\Collection::empty() method

edegaudenzi:
This solved #797

### DIFF
--- a/src/XeroPHP/Remote/Collection.php
+++ b/src/XeroPHP/Remote/Collection.php
@@ -18,6 +18,16 @@ class Collection extends \ArrayObject
     }
 
     /**
+     * Return whether or not the Collection is 0
+     *
+     * @return bool
+     */
+    public function empty()
+    {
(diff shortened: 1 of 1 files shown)


src/XeroPHP/Remote/Collection.php
```diff
@@ -17,6 +17,16 @@ public function addAssociatedObject($parent_property, Model $object)
...
     }
 
+    /**
+     * Return whether or not the Collection is 0
+     *
+     * @return bool
+     */
+    public function empty()
+    {
+        return !sizeof($this);
```
calcinai, line 27: Could you please consider using `count()` instead (`sizeof()` is an alias of `count()` and can be ambiguous due to its behaviour in other c-based languages)

edegaudenzi: Sure! Going to do that.

PR #802, opened 2021-04-28
Tweak README examples

### LINKED ISSUE #575: xero api php sdk 4 decimal places
Hi,
Using Xero in the past we have created invoices with a unit price of 6.667. This is due to the customer having 3 items @ 6.667 ea which provides an invoice total of $22.00 inc gst. (cash transaction, no silver)

We have created these (and other "odd" value) invoices manually via the Xero UI and using the csv file import process for batch creation of invoices, without issue.

To make our operation more efficient we have had a stand alone app created to manage many parts of our operation. One part of the app is to send the data to create the invoice to Xero, via the Xero API to create invoices (replaces the csv batch file process)

Whilst our app sends item values to 3 decimal places, the Xero API only accepts 2. This causes our invoices to be incorrect and worse when the customer payment is applied for the 3 decimal place value, xero rejects the payment.

Xero responded with:

Yes, by default Xero API uses 2 decimal points. However, by using a param ?unitdp=4, the API can support up to 4 decimal points.

Our programmer is also aware of:

https://stackoverflow.com/questions/27200459/rounding-in-xero-api-integration

however cannot pass 4 decimal places to Xero.

Can anyone please advise how if/how to set the API so it will receive and apply item values to 3 decimal places, same as Xero UI and CSV file import.

Our programmer is using PHP SDK

Thanks in advance.

Josh-G:
Split into commits:
* document how to set the unitdp parameter (resolves #575)
* remove link to the demo repo (no longer exists), and move the reference to `examples/` to the top of the example list
* add subsection headers to "Interacting with the API" to more clearly distinguish between examples

Let me know if you would prefer any tweaks, or if you would like to drop any of these changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,12 +115,13 @@ in the `XeroPHP\Models\Accounting` namespace. Additionally, there are models for
 
 Refer to the [examples](examples) for more complex usage and nested/related objects.
 
+### Instantiate an Application
 Create a XeroPHP instance (sample config included):
 ```php
 $xero = new \XeroPHP\Application($accessToken, $tenantId);
 ```
 
-Load a collection of objects and loop through them
+### Load a collection
 ```php
 $contacts = $xero->load(Contact::class)->execute();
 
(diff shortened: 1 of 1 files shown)


calcinai:
Thanks @Josh-G

PR #808, opened 2021-06-28
Adding default_content_type config value

matthewswain:
This PR adds a new item to `$_config_defaults`, to allow users of the library to override the default content type (`text/xml`) used for `Accept` header on HTTP requests.

This might seem like an odd request because the library handles parsing of responses, but until recently the Xero API returned JSON for many calls even when sending requests with `Accept: text/xml` meaning that where we previously consumed JSON we're now consuming XML.


calcinai:
Makes sense (the change, not Xero's behaviour)

Thanks!

PR #809, opened 2021-06-29
Fix use of config array when accessing default_content_type value

matthewswain:
This PR fixes a bug introduced in #808 where the Application config was accessed incorrectly.


calcinai:
@matthewswain thanks. I probably should have picked that up..

PR #810, opened 2021-07-01
Removes PUT from supported methods TaxRate

aguvillalba:
Removes PUT method from supported methods TaxRate because TaxRates in Xero must use POST for update and can also be created using POST. If PUT is supported then TaxRate cannot be updated since it has no GUID.


calcinai:
👍 thanks

aguvillalba:
Thanks to you for this great work! Let's keep it up!

PR #812, opened 2021-07-15
Add "Details" for Payment when used as a BatchPayment

mogilvie:
For consideration, 

Batch payment files are created with a user defined "Details" text where the Xero user can enter a message to transmit with payment for display on the recipients bank statement.

image

https://developer.xero.com/documentation/api/accounting/batchpayments#get-batchpayments

image


calcinai:
Thanks Mark

PR #826, opened 2021-09-27
Initialise all properties when a new model is constructed

### LINKED ISSUE #652: Getters throw exception if property is empty
Trying to develop something where I need to update the PayTemplate of an Employee in PayrollAU. 

If the employee doesn't have a PayTemplate set already, calling getPayTemplate throws an Exception "Undefined index: PayTemplate". I assume this would happen with other properties as well.

**Possible Solutions:**
Set array properties to an empty array on initialisation of the object
Return the array property, or a new empty array
Add "hasX" methods that return a boolean, e.g. $employee->hasPayTemplate()

glennschmidt:
Set collection properties to an empty collection, other properties to null.
Prevents an ‘Undefined index’ notice when calling getters.

I've addressed this via the constructor of the base model class, so it doesn't matter if some of the getters already check for  `isset` and others don't. Let me know if you think this approach will do the job, or if it will create other issues.

Resolves #652

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -76,6 +76,14 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
         $this->_dirty = [];
         $this->_data = [];
         $this->_associated_objects = [];
+
+        foreach (static::getProperties() as $property => $meta) {
+            if ($meta[self::KEY_IS_ARRAY]) {
+                $this->_data[$property] = new Collection();
+            } else {
+                $this->_data[$property] = null;
+            }
+        }
     }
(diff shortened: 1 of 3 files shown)


Healyhatman:
The problem is when you send the data to Xero, for some endpoints if a property is missing it ignores it, but if a property is NULL, it might delete it or otherwise perform an unwanted update. For example if all the properties are initialised to null and you update an Employee, any attributes not specified will be deleted.

glennschmidt:
I thought that's why we have the `$_dirty` list. Doesn't it only send properties that have been explictly set (marked as dirty)?

calcinai:
@glennschmidt exactly.  I think this should be fine

PR #837, opened 2021-12-16
Additional get/set methods for TrackingCategory and Payment.

hillsinfoodle:
First attempt at a pull request, open to alternative solutions, we require access to these properties and may just be doing things incorrectly, but seems possible these just were not required by anyone earlier so not added.


calcinai:
thanks @hillsinfoodle

PR #844, opened 2022-02-02
feat: Support subscriptions

### LINKED ISSUE #843: Fully Support SUBSCRIPTION webhooks
It would be nice to support subscription webhooks. Of course xero-php handles it by default, but some methods do not include it, for example:

- `Event#getEventClass()` - Only returns contact or invoice

There might be others too, but i've only noticed the above

ebebbington:
Closes #843 

I've done everything I can, so feel free to make your own comments, for example where to place these new files (I just placed them in `Models` for the time being).

Would be great to see this get passed, so my team can start utilising it for a project as opposed to adding this support ourselves into our own projects codebase

### DIFF
--- a/src/XeroPHP/Models/Plan.php
+++ b/src/XeroPHP/Models/Plan.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace XeroPHP\Models;
+
+use XeroPHP\Remote\Model;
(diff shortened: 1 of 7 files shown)


calcinai:
Looks great @ebebbington, thanks

Issue #840, opened 2022-01-20
PHP 8.1 - Return type issues


jsandfordhughes:
I am using this package with PHP 8.1 and while everything works correctly there are a few warnings that popup for interface compatibility.

The ones I have seen are the following:

`Return type of XeroPHP\Remote\Model::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed` 
`Return type of XeroPHP\Remote\Model::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool`
`Return type of XeroPHP\Remote\Model::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed`
`Return type of XeroPHP\Remote\Model::offsetSet($offset, $value) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void`
`Return type of XeroPHP\Remote\Model::offsetUnset($offset) should either be compatible with ArrayAccess::offsetUnset(mixed $offset): void`

Happy to do a PR for these if nothing is in the pipeline?

calcinai:
Yeah that would be great, thanks!

PR #848, opened 2022-03-21
PHP 8.1 support

p4veI:
Hello

In this MR I'm addressing Issue #840 & Switching required php version to 8.0 (7.4 support ends this year, would be nice to move on)

WDYT?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "homepage": "https://github.com/calcinai/xero-php",
   "license": "MIT",
   "require": {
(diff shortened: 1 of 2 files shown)


p4veI:
On second thought adding `#[\ReturnTypeWillChange]` on the methods that produce warnings in php 8.1 will not cause any BCs and can be added safely with support for older php releases. Updated.

calcinai:
@p4veI you're right, probably safer for now!

PR #849, opened 2022-04-15
Add Company Number as a field to the contact

webstik1811:


calcinai:
This is done in #850 sorry.

PR #866, opened 2022-08-05
Update EarningsRate.php

Healyhatman:
Fix invalid property type definition for EarningsRate => AccrueLeave

Currently, if an EarningRate has AccrueLeave checked then data is hydrated like this
```
        "Name" => "24 hour care - payment per shift"
        // Other stuff
        "AccrueLeave" => 0.0
      ]
```
instead of true.


calcinai:
👍

Healyhatman:
Can I get a release please, need to update clients :)

PR #870, opened 2022-09-16
Update Model.php

Healyhatman:
Add a static constructor so can more cleanly chain after instantiation, like this

`$rate = EarningsRate::make($application)->setEarningsType(...)->someOtherThing(...);`

Remove the `EARNINGTYPE_FIXED` rate because it's definitely not valid in Xero anymore. If you don't think that's the right call let me know.

### DIFF
--- a/src/XeroPHP/Remote/Model.php
+++ b/src/XeroPHP/Remote/Model.php
@@ -77,6 +77,11 @@ abstract class Model implements ObjectInterface, \JsonSerializable, \ArrayAccess
         $this->_data = [];
         $this->_associated_objects = [];
     }
+    
+    public static function make(Application $application = null)
+    {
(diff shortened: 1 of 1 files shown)


calcinai:
@Healyhatman are you doing this to match style with a particular framework? I'm just wondering about which verb is mose appropriate (I see `::create()` fairly regularly.  I assume you're just trying to avoid 

```php
$rate = (new EarningsRate($application))->setEarningsType(...)->someOtherThing(...)
```

Which I also see a bit.

Healyhatman:
It matches Laravel, where create() would immediately store a model in the database but make would just instantiate it

Healyhatman:
@calcinai

calcinai:
Agreed

PR #872, opened 2022-10-02
Add in the PUT type to create new TaxRates

mogilvie:
Was not able to create new tax types (although this used to work sometime ago) Added in the PUT type as per the documentation and now my original code works again.

API link here. https://developer.xero.com/documentation/api/accounting/taxrates

### DIFF
--- a/src/XeroPHP/Models/Accounting/TaxRate.php
+++ b/src/XeroPHP/Models/Accounting/TaxRate.php
@@ -132,6 +132,7 @@ class TaxRate extends Remote\Model
         return [
             Remote\Request::METHOD_GET,
             Remote\Request::METHOD_POST,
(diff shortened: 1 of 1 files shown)


mogilvie:
Note that the PUT method was removed in https://github.com/calcinai/xero-php/commit/cbd03cfd268e30735b1246b62d9639fbab1a9776 because the POST is _supposed_ to work for creating tax rates.

However, when you post the same content to Xero to tries to modify an existing tax rate, even when the names don't match. And if the number of TaxComponents in the existing (but differently named) TaxRate don't match the submitted TaxComponent count, then you get the error below.

POST not working:
image

Using PUT works:
image

calcinai:
LGTM, this kind of thing has been really inconsistent in the API over the years

PR #876, opened 2022-11-07
add PayrollUK/Leave class

pjw345:


pjw345:
Needs to be a subclass of Employee

PR #878, opened 2022-11-08
add getLeave to Employee class

pjw345:


pjw345:
added use TitleCaseKeysBeforeSave to Leave and Leave\Period classes

pjw345:
updated getter/setter and added LeaveType class

pjw345:
Add getStatutoryLeaveSummary to Employee class

pjw345:
Updated property 'lines' on class Timesheet should be 'timesheetLines'

PR #891, opened 2023-03-12
Add where date function

Healyhatman:
Fix a test namespace
Add the whereDate function
Nothing else


bretto36 (approved):

calcinai:
Hi @Healyhatman do you think this should just be the usual where method, and check if the parameter passed is a DateTime?

Healyhatman:
That'd probably be a good idea sure

PR #893, opened 2023-04-02
Normalise getIsPrimary method so it behaves the same as the other "Boolean"

bretto36:


composer.json
```diff
@@ -5,15 +5,15 @@
   "homepage": "https://github.com/calcinai/xero-php",
   "license": "MIT",
   "require": {
-    "php": ">=5.5.0",
+    "php": ">=8.0.0",
```
bretto36, line 8: Thought we should probably move to a php version that's even close to being supported - https://www.php.net/supported-versions.php
From memory someone got the package ready for 8.1. 
To try and run tests originally i had to scour the internet for a PHP 5 windows installer.

phpunit.xml.dist
```diff
@@ -8,7 +8,6 @@
          convertWarningsToExceptions="true"
          processIsolation="false"
          stopOnFailure="false"
-         syntaxCheck="false"
```
bretto36, line 11: doesn't do anything anymore so removed

phpunit.xml.dist
```diff
@@ -24,7 +23,7 @@
         </whitelist>
     </filter>
     <logging>
-        <log type="junit" target="build/logs/junit/phpunit.xml" logIncompleteSkipped="false"/>
+        <log type="junit" target="build/logs/junit/phpunit.xml"/>
```
bretto36, line 26: doesn't do anything anymore so removed

tests/Webhook/EventTest.php
```diff
@@ -106,7 +105,7 @@ public function testGetEventCategory()
...
...8-02-09T09:18:28.917Z","eventType":"UPDATE","eventCategory":"INVOICE","tenantId":"e629a03c-7ffe-4913-bd94-ff2fdb36a702","tenantType":"ORGANISATION"},{"resourceUrl":"https://api.xero.com/api.xro/2.0/Invoices/44aa0707-f718-4f1c-8d53-f2da9ca59533","resourceId":"44aa0707-f718-4f1c-8d53-f2da9ca59533","eventDateUtc":"2018-02-09T09:18:28.917Z","eventType":"UPDATE","eventCategory":"CONTACT","tenantId":"e629a03c-7ffe-4913-bd94-ff2fdb36a702","tenantType":"ORGANISATION"},{"resourceUrl":"https://api.xero.com/api.xro/2.0/Subscriptions/44aa0707-f718-4f1c-8d53-f2da9ca59534","resourceId":"44aa0707-f718-4f1c-8d53-f2da9ca59534","eventDateUtc":"2018-02-09T09:18:28.917Z","eventType":"CREATE","eventCategory":"SUBSCRIPTION","tenantId":"e629a03c-7ffe-4913-bd94-ff2fdb36a702","tenantType":"ORGANISATION"},{"resourceUrl":"https://api.xero.com/api.xro/2.0/Invoices/44aa0707-f718-4f1c-8d53-f2da9ca59533","resourceId":"44aa0707-f718-4f1c-8d53-f2da9ca59533","eventDateUtc":"2018-02-09T09:18:28.917Z","eventType":"UPDATE","eventCategory":"FUTURE_ENDPOINT","tenantId":"e629a03c-7ffe-4913-bd94-ff2fdb36a702","tenantType":"ORGANISATION"}],"firstEventSequence": 2,"lastEventSequence": 3, "entropy": "GATSEZXWIBPBRNQOTMOH"}';
```
bretto36, line 108: Someone had incorrectly updated the tests so this one was no longer passing

composer.json
Healyhatman: I was _going_ to get it ready for 8.1 but it hasn't happened yet so I'd skip this one for now

bretto36: @Healyhatman You want me to revert the composer changes or leave them as is?

Healyhatman: It's @calcinai 's repo but I expect he wouldn't want a major breaking change like that ? I mean anyone still on 5.5 deserves what they get but still....

calcinai: Hey team, I've been investigating how to do this gracefully, and I think composer is smart enough to resolve dependencies back to a previous supported version.  I think @Healyhatman's changes can go through on main, and we can make a branch to support PHP5.

bretto36: @calcinai  Agree, composer will install the earliest possible version that's compatible. Only reason i did this is because i couldn't get tests to run without doing a crazy amount of messing around to get a php 5 to install locally. 

https://stitcher.io/blog/php-version-stats-january-2023 - PHP 8+ is over 50% of usage now which is great for the ecosystem. According to this site 7.0 and below each have less than 1% representation.

bretto36:
I have removed the PHP 8 changes and just stuck with the underlying change + the namespace changes

Issue #806, opened 2021-05-12
Get headers from response object


jasonlewis:
I think I might be missing something here but there doesn't seem to be a way to get headers off of the response. I want to be able to preemptively check the rate limiting headers. Perhaps there's another way?

calcinai:
@jasonlewis does Xero return something before you hit the limit? There’s going to be a RateLimitExceeded exception thrown when you actually hit it.

jasonlewis:
Yeah Xero returns remaining limits in response headers so you can see what your limits currently are.

Taken the following snippet from here: https://developer.xero.com/documentation/oauth2/limits

> Each API response you receive will include the X-DayLimit-Remaining, X-MinLimit-Remaining and X-AppMinLimit-Remaining headers telling you the number of remaining against each limit.

calcinai:
Now you mention it, I do recall reading this somewhere.  If you create the request/response manually, you can get the raw response object, but that's not much help for normal usage.

If you use `$xero->save($model)`, this should return a response object that holds the headers... I think.

victormacko:
I've also come across this need, where can't pre-emptively view & slow down the API calls due to having no visibility of the returned headers for rate-limits.

The # remaining per min is usually the one I find being hit the most. Is there anything I can help with in terms of progressing this further?

Josh-G:
Are you able to accomplish the desired behaviour using a Guzzle middleware/handler? You can then pass this in using `setTransport`. We use this for retrying transitive errors with the API, and for handling rate limits.

victormacko:
Maybe by registering some event-listeners, and potentially adding some delays (to minimise the 'wait-and-retry' exception from being thrown) in the 'before' listener for guzzle too, although I think it'd be better to handle this within the Xero package rather than having to code it separately for each separate app.

calcinai:
@victormacko Happy for you to propose a way to handle this one

mogilvie:
Looking at it now, considering the following...

The XeroPHP\Remote\Query execute function gets the full response back from Guzzle, which contains the following:
![image](https://user-images.githubusercontent.com/7894754/203449524-93cbb59e-0404-4b45-8f01-dffff55cd2e1.png)

The Query object also contains the Application, so we could pass the limits back to the Application for use outside the bundle. Which kind of makes sense, Application keeps the record of transaction limits remaining during the life cycle of the action.

@calcinai Shall I work along these lines?

calcinai:
[...]

PR #897, opened 2023-08-04
Feed back API rate limits requests to the application. Demonstrate how to handle rate limit exceptions with Guzzle Middleware

mogilvie:
PR for issue #806

Pass the API Request Rate Limits back up to the Application to allow the user of the bundle to decide if they throttle api requests.

Have also added an option to throttle requests from within the bundle itself.

Set a configurable variable in the Application, which sets the maximum wait time that the application should delay if it hits an API limit. (Can be set false if no intention to retry requests)

Catch Xero 429 Response issues where the API limit has been reached.

Get the 'Retry-After' time from the Xero response header.

If the Retry-After period is less than the permitted maximum duration in the config, then sleep for the retry period, and retry the request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,12 +293,114 @@ See: [Signature documentation](https://developer.xero.com/documentation/webhooks
 Your request to Xero may cause an error which you will want to handle. You might run into errors such as:
 
 - `HTTP 400 Bad Request` by sending invalid data, like a malformed email address.
+- `HTTP 429 Too Many Requests` by hitting the API to quickly in a short period of time.
 - `HTTP 503 Rate Limit Exceeded` by hitting the API to quickly in a short period of time.
 - `HTTP 400 Bad Request` by requesting a resource that does not exist.
 
 These are just a couple of examples and you should read the official documentation to find out more about the possible errors.
 
-### Thrown exceptions
+### Rate Limit Exceptions
+
+Xero returns headers indicating the number for calls remaining befpre reaching their API lmits.
+https://developer.xero.com/documentation/guides/oauth2/limits/
+
+The Application is updated following every request, so that you can track the number of requests remaining until hitting the Xero limit using the Application::getAppRateLimits() method. It returns an array with the following keys and associated integer values.
+
+    'last-api-call' // The int timestamp of the last request made to the Xero API
+    'app-min-limit-remaining' // The number of requests remaining for the application as a whole in the current minute, limit is 10,000.
+    'tenant-day-limit-remaining' // The number of requests remaining for the individual tenant the minute, limit is 5,000.
+    'tenant-min-limit-remaining' // The number of requests remaining for the individual tenant the day, limit is 60.
+
+This can be used to decide if additional requests will throttled or sent to some message queue. For example:
+
+``` php
+    // If you know the number of API calls that you intend to make. 
+    $myExpectedApiCalls = 50;
+
+    // Before executing a statement, you could call the rate limits.
+    $apiCallsRemaining = $xero->getAppRateLimits();
+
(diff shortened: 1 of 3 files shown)


calcinai:
@mogilvie just reading through this–I'm wondering if this is the correct place to be handling the throttling.  I can see that you've got to be intentional with that config parameter, but potentially just sleeping in execution of the request could lead to some potentially undesirable timeouts, especially as the rest of the application will have no idea that it's caught/sleeping for that reason.

Is there a way we could move that handling out of the middle of the request while keeping it simple? For instance, you could inject a middleware into the Guzzle constructor and leverage the "retry" middleware?  See https://github.com/guzzle/guzzle/blob/7.7/src/RetryMiddleware.php

mogilvie:
Thanks @calcinai, wasn't aware of the middleware option. I'll do that.

mogilvie:
So, looking further, we can achieve the retry using the current Application::setTransport, where the parent bundle can construct a Client with the the RetryMiddleware in the HanderStack.

Shall I just document this, instead of modifying the Application?

mogilvie:
Have updated PR to allow user to skip the default guzzle client, and provide their own Client as a transport.
Added example and explanation in ReadMe.

calcinai:
@mogilvie Yeah that's what I was getting at! Sorry I didn't quite get here until you'd already written the code..

mogilvie:
I think I'm done. Review at your leisure.

calcinai:
@mogilvie I think this way is much cheaper, thanks for your work.  The only thing I think would improve would be to either extend or duplicate the guzzle middleware to catch the errors and increment those counters. It'd also remove the need to do all the plumbing in user/app space.  

Functionally I think it's great as is, so this would really just shuffling some code around. Thoughts?

calcinai:
Cool, well let's stick with how it is.  The thing that got me thinking about it was really just the `getAppRateLimits()` method where properties are transposed into an associative array, which is a nightmare for any sort of static analysis.  I didn't want to suggest doing something differently there if another approach would remove that anywhere.

mogilvie:
I could change that getAppRateLimits() array response to individual getters?

calcinai:
@mogilvie If you don't mind, I think that would be great.  Really appreciate the back and forward on this.

mogilvie:
swapped out getAppRateLimits for individual getters.

JamesPoel:
```
$application = new Application(
            $token,
            $tennantId ,
            constructClient: false
        );

        $handlerStack = GuzzleHttp\HandlerStack::create();

        // Add token refresh middleware
        $handlerStack->push(function (callable $handler) use ($config) {
            return function (GuzzleHttp\Psr7\Request $request, array $options) use ($config, $handler) {
                // Handle rate limit.
                $retryIfRateLimited = function ($request, $options, $handler) use (&$retryIfRateLimited) {
                    return $handler($request, $options)->then(
                        function ($response) use ($request, $options, $handler, $retryIfRateLimited) {
                            if ($response->getStatusCode() === 429) {
                                sleep((int) $response->getHeader('Retry-After')[0]);

                                return $retryIfRateLimited($request, $options, $handler);
                            }

                            return $response;
                        }
                    );
                };

                return $retryIfRateLimited($request, $options, $handler);
            };
        }, 'xero.pre_request');

        $application->setTransport(
            new GuzzleHttp\Client([
                'headers' => [
                    'User-Agent' => sprintf(Application::USER_AGENT_STRING, Helpers::getPackageVersion()),
                    'Authorization' => sprintf('Bearer %s', $token),
                    'Xero-tenant-id' => $tennantId,
                ],
                'handler' => $handlerStack,
            ])
        );
```

This is my solution, which is pretty elegant.

PR #899, opened 2023-08-11
Update Contact.php

julianmourell:
If we load from xero, we can only ever add new contact persons, I need a way to remove the old contact persons after having loaded a contact. This will accomplish that.


calcinai:
Cheers for the update.  Just our of curiosity, will the `->removeAll()` method on the collection not work in your use case?

julianmourell:
Yes that will solve the problem, didn't realise that existed. Thanks.

PR #900, opened 2023-08-16
Added support for MergedToContactID.

JackGreen411:
<b>The Scenario</b>

Whenever a contact has been merged into another, we received a webhook event of an archived contact, but no additional information. We then had to perform a GET request by its 'resourceId' to get the contact.

The 'Before the addition of MergedToContactID:' screenshot below, show the payload we received when we made the GET request, but no 'MergedToContactID' was present.

The important aspect of this PR, is ensuring we received the 'MergedToContactID' value, allowing us to update our contact data with the newly updated ContactID.

<b>Before the addition of MergedToContactID:</b>

image

<b>After the addition of MergedToContactID:</b>

image


elliotlings:
Bump. It would be good if MergedToContactID was returned with a Contact which it currently isn't.

calcinai:
LGTM, thanks

Issue #903, opened 2023-11-27
Setting null value to a property that is already null causes it to be "dirty"


bretto36:
I've found an issue where i have a Contact that is being returned with a null last name. 

If i write $contact->setLastName(null) it marks it as dirty even though the values are the same. The function at the bottom is the problem. 

Would it be feasible to change this to allow it to compare the value more accurately? I'm happy to do the PR, just want the feedback before i do the work

something like 

```
$currentValue = isset($this->_data[$property]) ? $this->_data[$property] : null;

if ($currentValue !== $value) {
```

Original Function
```
protected function propertyUpdated($property, $value)
    {
        if (! isset($this->_data[$property]) || $this->_data[$property] !== $value) {
            //If this object can update itself, set its own dirty flag, otherwise, set its parent's.
            if (count(array_intersect($this::getSupportedMethods(), [Request::METHOD_PUT, Request::METHOD_POST])) > 0) {
                //Object can update itself
                $this->setDirty($property);
            } else {
                //Object can't update itself, so tell its parents
                foreach ($this->_associated_objects as $parent_property => $object) {
                    $object->setDirty($parent_property);
                }
            }
        }
    }
```

PR #904, opened 2023-12-04
Handle null values

bretto36:
#903 For this issue here is the PR. 

This allows you to set a value to null and it does NOT trigger the dirty if the value was already null


bretto36:
@calcinai Thanks for responding on the issue, here is the fix. I can't run tests locally due to my local setup.

calcinai:
LGTM, thanks for this.  I don't see any reason this won't be backward-compatible, as I'd also consider the current behaviour a bug.

PR #905, opened 2023-12-12
Added Update Method to Application

dipaksarkar:
I have added update method to updating the model when GUID property doesn't exist i.e TaxRate

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -190,7 +190,7 @@ class Application
         /** @var Remote\Model $class */
         $class = $this->validateModelClass($model);
 
-        if(!$guid){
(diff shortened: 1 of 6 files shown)


calcinai:
Hi @dipaksarkar- can you please check your whitespace changes here? They detract from the content of the actual change.

It also look like that method is a copy+paste from elsewhere.  If the method is the same, can you please consider reusing the code.

dipaksarkar:
@calcinai Aah, It's happen due to my vscode auto format. Please check now I have resolved the issue.

dipaksarkar:
@calcinai I have checked but couldn't found a better way to updating existing methods rather than creating a new method.

PR #906, opened 2023-12-15
Add AccountId field to Invoice Line Items

oatmael:
As part of recent Xero API changes, `AccountId` is now a valid alternative to `AccountCode`. They've helpfully only announced this via Email, so I can't link to a good source on the extent of other changes, but you can see that the API docs have been updated to reflect this:

[Xero Invoice Line Items](https://developer.xero.com/documentation/api/accounting/invoices#creating-updating-and-deleting-line-items-when-updating-invoices)


calcinai:
thanks @oatmael

PR #911, opened 2024-04-19
fix credit note allocation error

waeltech:


calcinai:
@waeltech thanks for the contribution. Can you please provide some detail about these changes in the PR?

waeltech:
Hi @calcinai,  when you try to save Credit Note to allocate against invoices, it was sending all allocations including the existing one, if you have credit note with existing allocations and you want to set only new allocation and leave previous one as is.

calcinai:
Thanks @waeltech, although I'm not completely sure what you're describing.  As far as I can see, there'd be three cases:

1. There are no allocations and you need to add one
2. There are some allocations and you need to add one
3. There are some allocations and you want to replace them

If that's the case, these cases can be handled already by manipulating the collection in that attribute.  Can you perhaps give an example of the problem, and what this PR solves (in pseudocode if that's easier)

waeltech:
Hi @calcinai 

- You have a credit note issued for £1000.
- A customer makes a new purchase of £200.  
- You want to allocate a portion of the £1000 credit note to this new invoice of £200.
- if this is the first time to add allocation there is no problem at all.  * There are no allocations and you need to add one.
- Customer decide to make new purchase of £300. * There are some allocations and you need to add one
- using the same code seems to result in Xero API 400 error as it seems like The API trying to re-allocate the previous allocations.

The Problem:

The code seems to be adding past allocations on top of the new allocation you're trying to create. This results in multiple allocations being sent to Xero in a single request.

Why This Causes an Error:

Xero interprets the multiple allocations as trying to allocate the credit note amount twice. Since an invoice can typically only be allocated to once, this attempt to allocate to a previously paid invoice fails (400 error).

The Fix:
as suggested in the PR.

Screenshot 2024-04-24 at 08 31 32
 Screenshot 2024-04-24 at 08 41 51
Screenshot 2024-04-24 at 08 45 13

PR #912, opened 2024-05-03
Remove slash from base url

mogilvie:
The XeroPHP\Application\URL concatenates the Application config base_url with the path.

```
    /**
     * @return string
     */
    public function getFullURL()
    {
        $url = sprintf('%s/%s', $this->base_url, $this->path);

        return$url;
    }
```

resulting in a double slash after the base url. https://api.xero.com//api.xro/2.0/Journals?paymentsOnly=true&offset=33300.

Remove the trailing forward slash from the base_url.

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -15,7 +15,7 @@ class Application
 
     protected static $_config_defaults = [
         'xero' => [
-            'base_url' => 'https://api.xero.com/',
+            'base_url' => 'https://api.xero.com',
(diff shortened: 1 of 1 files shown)


calcinai:
@mogilvie Does this cause an issue with anything in particular? I know it's not correct, just can't think when/how it was introduced!

mogilvie:
No, just identified it while I was resolving a different issue.

calcinai:
I'm going to leave this for a new version, as it's possible people are relying on that trailing slash.

PR #913, opened 2024-05-03
Fix: Handle Root Warnings and Access the Responses from Queries

patrickmallen:
Hi.
@p4vel mentioned in #888 there is a bug when accessing the warnings at the root of the response, because of the change mentioned here: https://developer.xero.com/documentation/api/efficient-data-retrieval

However, I don't think their solution is in keeping with how the system currently works, and I think adding it to the list of elements is likely going to lead to some issues with how other users are parsing the data.
As such, I've added it as a private property under the responses (similar to the root error).
I've also added proper handling for the json response, which I don't believe their solution would have worked with.
I've added a small unit test for this and confirmed that it did work.

Given the documentation indicates that this comes up in the query section, I've also added a private property on the query to save the response during the execution of the query, and a public method for accessing it.
This should allow users to access the response and parse any warnings that may have come through, while handling the content of the response as they like (and shouldn't interrupt any current implementations).

While I was doing this I noticed some issues with the inconsistent PHP versioning.
For example: in the Application file the private attributes on lines 32 to 35 (e.g. $lastApiCall) are all typed properties, however typed properties are not supported until PHP 7.4 (which is inconsistent with the composer PHP version of >=5.5.0).

I wasn't aware of whether you would want to update the composer version or remove the conflicts, so I've left them be for now.
If you'd like, I can make another pull request with the requisite change, should you tell me.

### DIFF
--- a/src/XeroPHP/Remote/Query.php
+++ b/src/XeroPHP/Remote/Query.php
@@ -38,6 +38,9 @@ class Query
 
     private $params;
 
+    /** @var Response|null $response  */
+    private $response;
+
     public function __construct(Application $app)
     {
         $this->app = $app;
@@ -49,6 +52,7 @@ class Query
         $this->includeArchived = false;
         $this->createdByMyApp = false;
         $this->params = [];
+        $this->response = null;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


calcinai:
Looks good to me. I think there's value in exposing those as properties regardless, given how the API has changed over the years.

I think we need to bump to PHP 8 anyway TBH, but this is probably not the right place to do that.

p4veI:
I understand this fixed the issue I addressed in https://github.com/calcinai/xero-php/pull/888 ?

patrickmallen:
It should have done. Though in a slightly different way.

PR #918, opened 2024-07-01
Support the new PageSize\pagination property

bretto36:
No longer do they make it 100 default per page. 

I wanted to be able to open up the getPageInfo() method, but ->execute() just gets the results, so you can't actually know what those values were in the <PageInfo> element with the current implementation. 

Are there any plans to move this to a newer minimum version of PHP and introduce some newer features?

Switching to json only could save some filesize in the transfers going forward too. I'm happy to contribute to the package but if no plans for any improvements i won't push the envelope


src/XeroPHP/Remote/Response.php
```diff
@@ -318,6 +318,7 @@ public function parseXML()
         foreach ($sxml as $child_index => $root_child) {
             switch ($child_index) {
                 case 'PageInfo':
+                case 'pagination':
```
bretto36, line 321: @calcinai  So Xero have just done a new release and it's screwed everything up. Instead of PageInfo they've changed it to pagination

"pagination": {
        "page": 1,
        "pageSize": 100,
        "pageCount": 42,
        "itemCount": 4112
    },

<pagination>
        <page>1</page>
        <pageSize>100</pageSize>
        <pageCount>42</pageCount>
        <itemCount>4113</itemCount>
    </pagination>
    
    and using the pageSize parameter works again
?pageSize=1000
<pagination>
        <page>1</page>
        <pageSize>1000</pageSize>
        <pageCount>5</pageCount>
        <itemCount>4113</itemCount>
    </pagination>

bretto36:
I've changed this PR to not be about the PageInfo thing anymore, Its merely some new fields and the new pagination response now too

lewiswilbynm (approved):

src/XeroPHP/Remote/Response.php
lewiswilbynm: I can confirm that Xero has again changed their API and that the new pagination value is now handled (I've tested that part).

I have not tested anything relating to the practice manager files but this also looks OK and so does the page size changes.

bretto36:
Xero just don’t even send out any information about this stuff. Told us they were releasing the change but never actual described the change. Then rolled it back after a day and never let anyone know. Then again they release ANOTHER change that they didn’t give any warning for.
And then have the nerve to charge developers for use of the api which they never make improvements to. 😡😡😡

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

oatmael (approved):

src/XeroPHP/Remote/Response.php
oatmael: confirmed as well. really not happy with Xero on this one

lewiswilbynm: @calcinai any chance of merging and tagging this?

lewiswilbynm:
@calcinai any chance you can tag it too?

calcinai:
v3.7.0..

PR #921, opened 2024-07-02
PHP 8.1 

bretto36:


src/XeroPHP/Models/Accounting/Account.php
bretto36, line 1: @calcinai  I'd love your opinion on a few things in this PR if you have the time. 

We could move the method calls to magic methods which would help to clean up the code considerably. 

We still just manage the `getProperties` function, and any basic get/set we can completely ditch and just add an @method for 

See Remote/Model change (very bottom of file)

bretto36:
Got all the tests working again. Added some extras. Updated to php 8.1, added alot of return types and removed superfluous params in PHPDocs

bretto36:
@calcinai  - Are we able to make a decision and move this package up some PHP versions please, maybe even a new major version of the package so we don't affect the poor people stuck on PHP 5 and 7. 

I have a security advisory of my fork - https://github.com/feesynergy/xero-php/security/dependabot/1 i don't want to fix it just on my fork

I'd love to move the project along so we can merge in the Tests PR #945  that i created on this package too. Would love to have the package using a modern php and some modern principles, looks like a few other PR's that are potentially mergable that clean the codebase up

If you don't plan on updating anything please let me know and i'll just start using my fork and not worrying about the parent package anymore. 

Thank you for your time!

bretto36:
I'm not saying we have to merge in this PR, but i would at least like the composer.json to reflect a reasonable PHP version.

Issue #923, opened 2024-07-18
Call to Invoices returns four null invoices - may be related to new invoicing released by Xero


gurubobnz:
We noticed that we were getting some null object Invoices coming back from the Xero API, specifically:

```
	$xero = Load::lib('xero')->getApplication();
	$invoices = $xero->load(Invoice::class)
		->where('Type', 'ACCREC')
		->page(1)
		->execute();
	dd($invoices);
```

Returns not 100 results (the number on the page), but 104. The extra four appear to be all blank. We started seeing this on one of our tenants as of 17/07/2024 at approx 11:54am NZST. We confirmed that it happening on another four tenants also, with differing versions of this library.

At a glance the Xero API explorer appears to not be reporting the same, but I am confirming this.

Here is the dump that we get from the above code:

```
XeroPHP\Remote\Collection {#139
  #_associated_objects: null
  storage: array:104 [
    0 => XeroPHP\Models\Accounting\Invoice {#136
      #_data: array:32 [
        "Type" => null
        "Contact" => null
        "LineItems" => null
        "Date" => null
        "DueDate" => null
        "LineAmountTypes" => null
        "InvoiceNumber" => null
        "Reference" => null
        "BrandingThemeID" => null
        "Url" => null
        "CurrencyCode" => null
        "CurrencyRate" => null
        "Status" => null
        "SentToContact" => null
        "ExpectedPaymentDate" => null
        "PlannedPaymentDate" => null
        "SubTotal" => null
        "TotalTax" => null
        "Total" => null
        "TotalDiscount" => null
        "InvoiceID" => null
        "RepeatingInvoiceID" => null
        "HasAttachments" => null
        "Payments" => null
        "Prepayments" => null
        "Overpayments" => null
        "AmountDue" => null
        "AmountPaid" => null
        "FullyPaidOnDate" => null
        "AmountCredited" => null
        "UpdatedDateUTC" => null
        "CreditNotes" => null
      ]
      #_dirty: []
      #_associated_objects: []
      #_application: XeroPHP\Application {#108
        #config: array:1 [ …1]
        -transport: GuzzleHttp\Client {#119 …1}
      }
    }
    1 => XeroPHP\Models\Accounting\Invoice {#127
      #_data: array:32 [
        "Type" => null
        "Contact" => null
        "LineItems" => null
        "Date" => null
        "DueDate" => null
```

and so on, for the first four records, then we get normal invoices.

[...]

PR #924, opened 2024-07-18
Update Response.php

gurubobnz:
Ignore pagination element in response, proposed fix for https://github.com/calcinai/xero-php/issues/923

### DIFF
--- a/src/XeroPHP/Remote/Response.php
+++ b/src/XeroPHP/Remote/Response.php
@@ -320,6 +320,9 @@ class Response
                 case 'PageInfo':
                     // TODO: We can potentially handle the page info and make it a value on the response object
                     break;
+                case 'pagination':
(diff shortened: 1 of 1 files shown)


oatmael:
@calcinai this change was also made in #918 :
https://github.com/calcinai/xero-php/pull/918/files#diff-d7be814166ab8f10b0e0f81580d74f84450e12e1cb79fd7b627d536b2572c33cR321

so there's now duplicate cases in the switch cases:
https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/Response.php#L324

thankfully this functionally doesn't matter, but try and keep in mind for future :)

calcinai:
Gosh my bad. It'll just break at the first, but I'll tidy up for next release.

PR #925, opened 2024-07-23
Remove PHP 7.1+ references to keep 5.6+ compatibility.

bryce-squirrel:
Removes updates that break compatibility with PHP 5.6

### DIFF
--- a/src/XeroPHP/Application.php
+++ b/src/XeroPHP/Application.php
@@ -44,7 +44,7 @@ class Application
      * @param $tenantId
      * $param $constructClient
      */
-    public function __construct($token, $tenantId, ?bool $constructClient = true)
+    public function __construct($token, $tenantId, $constructClient = true)
     {
         $this->config = static::$_config_defaults;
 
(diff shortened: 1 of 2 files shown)


philhawthorne:
Hey @calcinai 

Any chance you can push this out as a .1 fix release soon?

With the Xero changes, this is a blocker for us to update to the latest release which address those 😢

Thanks for all your work on this one

Healyhatman:
OR you could upgrade your PHP version to one that was supported in the last 10 years?

philhawthorne:
Not all of us get to inherit tech-debt free code. We can try give back to a project where we can though and help others that might also be in a similar position.

src/XeroPHP/Application.php
```diff
@@ -462,14 +462,14 @@ public function delete(Remote\Model $object)
         return $object;
     }
 
-    public function updateAppRateLimit(int $appMinLimitRemining)
+    public function updateAppRateLimit($appMinLimitRemining)
```
bretto36, line 465: You reckon you could add the phpdoc version of the prop types? Just so there is no loss in the IDE auto complete

/*
* @param int $appMinLimitRemining.
*/

bretto36:
@philhawthorne  still waiting for this to get rolled out, or did you have to do something different to get around it?

philhawthorne:
Hey @bretto36 

We ended up forking and installing via our fork in the end to get this out and around it

PR #933, opened 2025-02-19
Allow BadRequestException to contain the response

oatmael:
For debugging/logging purposes, it can be useful to access the raw response when dealing with `BadRequestException` and build an appropriate error message yourself.

Since the `Response` object throws when it's parsed, and `Query` / `Application` only return the response when successful, simply catching remote exceptions doesn't expose the request / response that caused it.

This PR adds a getter/setter on the `Remote/Exception` object for the `Response` that caused it, so caught exceptions can get access to this information:
```php
try {
  $application->save(...);
} catch (BadRequestException $e) {
  $response = $e->getResponse();
  if ($response) {
    logElementErrors($response->getElementErrors());
  }
  ...
}
```

For the sake of limiting breaking changes, I have abstained from adding the `Response` as a constructor option, and only setting the response on the `BadRequestException` (as that is where it's most useful).

Future PRs could re-consider how this is handled (i.e., adding a static `fromResponse` function to all remote exceptions).

### DIFF
--- a/src/XeroPHP/Remote/Exception.php
+++ b/src/XeroPHP/Remote/Exception.php
@@ -4,5 +4,15 @@ namespace XeroPHP\Remote;
 
 class Exception extends \XeroPHP\Exception
 {
-    //
+    protected ?Response $response = null;
+
+    public function setResponse(Response $response): self
+    {
+        $this->response = $response;
+        return $this;
+    }
+
+    public function getResponse(): ?Response {
+        return $this->response;
+    }
 }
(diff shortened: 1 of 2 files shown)


calcinai:
Makes sense, there are obviously lots of ways this can be achieved, but for the sake of simplicity, this is probably the best approach for now.

There was talk a while ago about adding hooks, which is probably somewhat already available if you manually construct your Guzzle client, although you'd miss the parsing.

Thanks @oatmael

PR #934, opened 2025-05-22
pass through response to NotFoundException

scrummitch:
Hi there, 

We've found in our application of the `LangleyFoxall\XeroLaravel` library that Xero often returns important information in the error response bodies that is important for us to store and present options for our users to solve their issue. 

This PR solves one major blocker for us which attaches the response body and parses it correctly from Xero to the NotFoundException so errors can be read out.

image

This lets us do things like this:
```php
try {
    $xero->payrollAUPayrollCalendars()->get();
} catch (Exception\NotFoundException $e) {
    dd(
        $e->response->getRootError(),
    );
}
```

and get out super helpful responses like:
```
array:3 [
  "code" => "0"
  "type" => "Error"
  "message" => "Payroll has not been provisioned"
]
```

### DIFF
--- a/src/XeroPHP/Remote/Request.php
+++ b/src/XeroPHP/Remote/Request.php
@@ -21,6 +21,8 @@ class Request
 
     const CONTENT_TYPE_XML = 'text/xml';
 
+    const CONTENT_TYPE_APPLICATION_XML = 'application/xml';
+
     const CONTENT_TYPE_JSON = 'application/json';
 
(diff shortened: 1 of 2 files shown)


calcinai:
Hi @scrummitch, check out #933 and its usage as it adds a `setResponse()` method to the base Exception, which I think you should use for consistency here.

scrummitch:
Thank you @calcinai for the quick feedback! I've integrated the changes and we now pass response to the exception via setResponse().

PR #939, opened 2025-11-13
FIX: PHP 8.4 implicitly nullable parameter declarations deprecation

dtokos:
FYI: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


dtokos:
@baci266 I have opened #941 for that.

PR #941, opened 2026-01-05
Add @return type for Model::offsetUnset()

dtokos:
This fixes the following deprecation:

```
User Deprecated: Method "ArrayAccess::offsetUnset()" might add "void" as a
native return type declaration in the future. Do the same in implementation
"XeroPHP\Remote\Model" now to avoid errors or add an explicit @return annotation
to suppress this message.
```

Reported by @baci266 in [#939](https://github.com/calcinai/xero-php/pull/939#issuecomment-3705930533)


baci266:
thanks @dtokos  :) 

@calcinai could you pls publish a new release with these fixes? 


As a thank you a I added workflow for phpunit https://github.com/calcinai/xero-php/pull/942

PR #942, opened 2026-01-05
feat: add PHPUnit workflow configuration

baci266:


baci266:
you can check actions at https://github.com/baci266/xero-php/actions

composer.json
```diff
@@ -5,15 +5,15 @@
   "homepage": "https://github.com/calcinai/xero-php",
   "license": "MIT",
   "require": {
-    "php": ">=5.5.0",
+    "php": ">=7.4",
```
baci266, line 8: I m not sure if this php update should be in this PR. 
But code is using typed properties (introduced in php 7.4). 

https://github.com/calcinai/xero-php/blob/master/src/XeroPHP/Remote/Exception.php

tests/Remote/ModelTest.php
```diff
@@ -209,7 +210,7 @@ public function testPracticeManagerClientListIsReturned()
 
         foreach ($model->getContacts() as $contact) {
             $this->assertEquals(220, $contact->getID());
-            $this->assertTrue($contact->getIsPrimary());
+            $this->assertSame('Yes', $contact->getIsPrimary());
```
baci266, line 213: this is expected?

calcinai:
We're targeting PHP 8 now, so it probably makes sense yo update to that too.

tests/Remote/ModelTest.php
bretto36: Yes this is what it should be. I wrote the XeroPracticeManager code and this was an oversight when i switched it from Bool

bretto36:
> We're targeting PHP 8 now, so it probably makes sense yo update to that too.
@calcinai 
If we are targetting 8 are we able to close out some of the outstanding PR's or merge in the ones that are 8 compatible. 

I feel like this is stuck in a bit of limbo at the moment

Issue #947, opened 2026-04-16
Error accessing  XeroPHP\Models\Files\Object


jny986:
When trying to access the class for File Objects (XeroPHP\Models\Files\Object) I get the following error 

```
Symfony\Component\ErrorHandler\Error\FatalError 

  Cannot use "Object" as a class name as it is reserved

  at vendor/calcinai/xero-php/src/XeroPHP/Models/Files/Object.php:5
      1▕ <?php
      2▕ 
      3▕ namespace XeroPHP\Models\Files;
      4▕ 
  ➜   5▕ class Object
      6▕ {
      7▕     const OBJECT_GROUP_ACCOUNT = 'Account';
      8▕ 
      9▕     const OBJECT_GROUP_BANKTRANSACTION = 'BankTransaction';

```

Working with latest version at time (v2.8.0)

To recreate try initing Class:
```
new XeroPHP\Models\Files\Object()
```

PR #948, opened 2026-04-16
Change Files\Object to Files\FileObject

jny986:
Fix for issue [Error accessing XeroPHP\Models\Files\Object](https://github.com/calcinai/xero-php/issues/947)

Rename of XeroPHP\Models\Files\Object to XeroPHP\Models\Files\FileObject to fix use of restricted word.

I have checked as best I can for references in the package for XeroPHP\Models\Files\Object.

Happy to here any feedback on how to do this better.

### DIFF
--- a/src/XeroPHP/Models/Files/FileObject.php
+++ b/src/XeroPHP/Models/Files/FileObject.php
@@ -2,7 +2,7 @@
 
 namespace XeroPHP\Models\Files;
 
-class Object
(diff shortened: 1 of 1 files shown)


calcinai:
Looks good to me, although I would put money on the fact that there will be applications checking for `instanceof Object` so this will need a release bump.  Who knows why I named it that to begin with...

